Issue #4, opened 2017-10-25
Drag and drop file crashes the program


SpaceOgre:
Version: 151
OS: Windows 10

When I drop files into the program from explorer it crashes. If I open the same file from with in the program it works fine.

Could not find a log file to give you.

zarunbal:
@SpaceOgre If the Popup from Windows was displayed that the application had an unexpected error the exception should be visible in the EventViewer from Windows (eventvwr.exe -> Windows Logs -> Application, and there the Application Error entry **and** the .NET Runtime exception)

SpaceOgre:
Application Error:
```
Faulting application name: LogExpert.exe, version: 1.5.0.0, time stamp: 0x59e92028
Faulting module name: msvcrt.dll, version: 7.0.15063.0, time stamp: 0x3280d1b7
Exception code: 0xc0000005
Fault offset: 0x0000000000073680
Faulting process id: 0x2d24
Faulting application start time: 0x01d34d7062663cbe
Faulting application path: C:\Users\DanielNordström\Downloads\v151\LogExpert.exe
Faulting module path: C:\WINDOWS\System32\msvcrt.dll
Report Id: 34db33ca-f641-42f8-a350-8b66f11888f3
Faulting package full name: 
Faulting package-relative application ID: 
```

.Net Error
```
Application: LogExpert.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.AccessViolationException
   at System.Windows.Forms.UnsafeNativeMethods.CallWindowProc(IntPtr, IntPtr, Int32, IntPtr, IntPtr)
   at System.Windows.Forms.NativeWindow.DefWndProc(System.Windows.Forms.Message ByRef)
   at System.Windows.Forms.ComboBox.DefChildWndProc(System.Windows.Forms.Message ByRef)
   at System.Windows.Forms.ComboBox.ChildWndProc(System.Windows.Forms.Message ByRef)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr, Int32, IntPtr, IntPtr)
```

zarunbal:
@SpaceOgre Thanks for the infos. I can reproduce the problem.

I have found a workaround, until I can fix the root of the problem (which will take some time). Could you please test the workaround if this works for you?

Create a start.cmd (you can name it differently) in the folder where the LogExpert executable is located with the content below. After you saved this start LogExpert with this CMD file.
```
[...]

PR #23, opened 2018-01-24
Remove unnecessary code.

louisledev:
Hi Zarunbal.
This is just an small step after your last cleanup. (I actually did the same as you and this solved the bug https://github.com/zarunbal/LogExpert/issues/4 imho)
When do you plan to do a new release since the above bug is very annoying for us?
Thanks.
Louis


zarunbal:
Hi Louis, 
thanks for your Pull Request, I really appreciate your help! I hope that I can create a new release at the weekend.
About #4 I have to test if your change fixes the bug on my machine too.

I have to improve the build script these build fails are annoying.

zarunbal:
New release is there. Was not the weekend but nearly.

PR #43, opened 2018-03-14
Development SFTP Plugin and minor enhancements

zarunbal:
Mainly a test of the Build script if it runs now with pull requests


zarunbal:
test failed ...

PR #63, opened 2018-05-17
Timestamp improvement

detlefs:
TimeStampColumnizer: added a way to ignore first character in line, if it is a bracket ((, {,[).
I have this case in many of our logs, so I added this. I think it should not have any negative effect on other logs.
This change will properly columnize lines like:
`[2018-03-27 16:23:40,343] [] Current database version is 9`


zarunbal:
Your changes are merged to Development. Thank you for your contribution!

PR #67, opened 2018-06-29
Handle lines which do not match regex

aliask:
If a line fails the Regex, don't just omit it - stick it in the last column.

I picked last column because I have many log files which have the last column being the "text" field, and the "text" sometimes contains newlines. I imagine I'm not the only one with this issue :)


zarunbal:
Merged your changes into Development (7218f593d9de00f39d380cf3851694354b23a78e). Thanks for your contribution!

PR #68, opened 2018-07-02
Apply updated regex on closing dialogue box

aliask:
Previously, updating the regex required restarting LogExpert in order to re-read the updated XML file. 
Configure() should update the active configuration after it has validated it.


zarunbal:
Merged your changes to Development (b7799501aa632665a75f045423a623bfa23cdd78). Github loses the reference if I change the destination branch :/ 
Thanks for your contribution!

PR #71, opened 2018-07-24
Development

SeniuuS:
I added compatibility with RenciSsh (as I needed an API which didn't have a free-trial)
Also I thought it would be better to specify the Chilkat Key in the config panel.
The user can choose between Renci and Chilkat in the config of SftpPlugin


zarunbal:
Thanks for your Pull Request. I have some questions
- Why the key as input field? In my view most users which uses the ssh columnizer doesn't have a license from chilkat. But if it is build with a valid license stream then everyone can use the columnizer without restrictions?
- What features has the renci implementation? As I have seen its licensed under MIT. Some more noteable features?
- I assume the other changes (Powershell Tail, ...) are not ment to be in this pull request?

SeniuuS:
Hello,

I thought the key was to activate Chilkat SSH API. So it is the user responsibility to have a valid key in order to not be blocked after 30 days of utilization. Someone who is only downloading LogExpert (not the source) will need to specify a Chilkat Key in order to use its SSH functionnalities, now he can specify it in the GUI instead of the source code.

As for Renci I added it to have an alternative for Chilkat. Chilkat will not be available after 30 days of use if you don't have a key. There is no noteable features between the two API. It is just to prevent this 30 days trial...

Yes the Powershell thing should not be in the pull request, I don't know why it added automatically as I pushed to my fork...

Thanks for your response

PR #77, opened 2018-09-10
Bug fixes #70: OK button in Highlighting and action trigger window

### LINKED ISSUE #70: OK button in Highlighting and action trigger window
OK button in Highlighting and action trigger window is unintuitive. 
I have often caught myself on entering new search string, choose colors and click OK. Then wondering why it doesn't work. Then realized - forgot about Add button.
No warnings, question and have to repeat again. 

Is it possible to add warning/question on OK button when new string has been add and not confirmed by Add button or just add automatically?

I've noticed that OK button works as expected when string has been changed. Missing the same behavior when add new string. 

LogExpert Version - 1.6.13

miltob:
Bug fixes.


zarunbal:
Thanks @miltob for your bugfix. I merged your change into the branch.

PR #91, opened 2019-02-14
More columnizer

miltob:


miltob:
@zarunbal There are some changes that I have done here.

1. Added an AutoColumnizer. In the future, I am thinking about add a "priority" field and a "DoesFit" method in ILogLineColumnizer interface so all columnizer can be loaded automatically.
2. Modified JsonColumnizer for better compatibility.
3. Added two new columnizers.

Please take a look and let me know your opinion. I'd like to make changes gradually. I'll try make smaller but frequent ones in the future.
P.S. I think the master is a bit behind and there are some good features are not in formal releases. Maybe we could release more often so that other people enjoy the latest changes.

miltob:
Supported Priority for AutoColumnizer.
When open a file, the AutoColumnizer should be able to find the most relevant columnizer for most of the case.

zarunbal:
Thank you @miltob for this new columnizer. I have currently a bit trouble with my PC and limited access to VS. I try as fast as possible to review and merge this.

zarunbal:
Thanks again for the PR and I would appreciate gradually changes.

PR #95, opened 2019-03-05
Added the ability to configure how many items in the filter will be displayed

Hirogen:
updated 3 files

- with pattern matching
- usings with streams instead of manuel close
- removed unnecessary object initialization
- expression bodies
- null coalescing
- string interpolation and verbatim strings

### DIFF
--- a/src/LogExpert/Config/ConfigManager.cs
+++ b/src/LogExpert/Config/ConfigManager.cs
@@ -126,6 +126,7 @@ namespace LogExpert
             lock (loadSaveLock)
             {
                 Settings settings;
+
                 if (fs == null)
                 {
                     settings = new Settings();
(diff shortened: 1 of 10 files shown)


Hirogen:
made a few optimizations

Hirogen:
Added the ability to configure how many filter items are
- listed default: 20, minium: 10, maximum: 30
- stored default 20: minimum: 10, maximum: 100

With this change a scrollbar will be displayed in the filter drop down

zarunbal:
Thanks for your changes. Thanks that you keep on contributing! You even translated code comments 👍 

Please next time try to keep the changes where needed (or something changed). Its hard to review so many changes. I know there are many parts in there which can be fixed automatically with e.g. Resharper and other tools (Its tempting to let it run free over the code base, I know). These automatic changes creates a really big PR which I have to review and search where are the real changes are.

Another thing is not everything these tools suggest are better! Especially with loggers and string interpolation. 
See by this example. The screenshots are from a decompiler

**Normal log string code**
` _logger.Info("No persistence data for {0} found.", FileName);`
Decompiled
2019-04-03 22_31_07-Window

**String interpolation code**
`_logger.Info($"No persistence data for {FileName} found.");`
Decompiled
2019-04-03 22_30_10-Window

So the string interpolation leads to a string concat every time even if the log level is not active. The Logger checks before the string.format execution if the log level is active.

<My own opinion. I don't want to offend someone or create endless discussions. I just leave this here>
In my opinion not every "code shortener" is better. I try to write code expressive as possible but not at every price to be short as possible. But most of the helper tools suggest to cut the code really short. But this leads to mostly one line codes which are really lengthy and in my opinion harder to read then e.g. a if over multiple lines with braces. A question mark between >100 characters is harder to spot then a multi line if

I know this code base has **some** places to **improve** ;)

Hirogen:
> 
> 
> Thanks for your changes. Thanks that you keep on contributing! You even translated code comments 👍
> 
> Please next time try to keep the changes where needed (or something changed). Its hard to review so many changes. I know there are many parts in there which can be fixed automatically with e.g. Resharper and other tools (Its tempting to let it run free over the code base, I know). These automatic changes creates a really big PR which I have to review and search where are the real changes are.
> 
> Another thing is not everything these tools suggest are better! Especially with loggers and string interpolation.
> See by this example. The screenshots are from a decompiler
> 
> **Normal log string code**
> ` _logger.Info("No persistence data for {0} found.", FileName);`
> Decompiled
> 2019-04-03 22_31_07-Window
> 
> **String interpolation code**
> `_logger.Info($"No persistence data for {FileName} found.");`
> Decompiled
> 2019-04-03 22_30_10-Window
> 
> So the string interpolation leads to a string concat every time even if the log level is not active. The Logger checks before the string.format execution if the log level is active.
> 
> <My own opinion. I don't want to offend someone or create endless discussions. I just leave this here>
> In my opinion not every "code shortener" is better. I try to write code expressive as possible but not at every price to be short as possible. But most of the helper tools suggest to cut the code really short. But this leads to mostly one line codes which are really lengthy and in my opinion harder to read then e.g. a if over multiple lines with braces. A question mark between >100 characters is harder to spot then a multi line if
> 
> I know this code base has **some** places to **improve** ;)

Ah thanks for the info, I did not know that interpolated strings are concatenated, I will keep that in mind!

For Future Changes, I will split code formatting changes and "real changes", and create more smaller commits, so changes are better traceable.

I will go through the German comments and translate them to English!

I use LogExpert daily its probably the best Log-Tool on windows! 💃

PR #96, opened 2019-03-20
Avoid duplicate loading of colors

igitur:
The designer file was changed a *lot* in 30c6de5a129c65e0c29ff70033109aeb1c420f98 . Part of this added code to load colors, but the colors are already loaded in `ColorComboBox`'s constructor. So you end up with 36 entries for each color combo box instead of 18. 

(By the way, I'm assuming we should base PRs against the `Development` branch?)


zarunbal:
You are right I can reproduce this error, thanks for the fix! Yes base it against the `Development`

PR #97, opened 2019-03-20
Set correct combobox index for known colors

igitur:
Known colors currently show as "Custom", because when setting .SelectedItem, the first entry (index 0) is a match.

This PR changes it so that if a known color is in the last, the last item is used, else the old behaviour continues.

Old behaviour:
image

New behaviour:
image


zarunbal:
Thanks again for your PR :)

PR #98, opened 2019-03-20
Use date format from regional settings

igitur:
Implements simple date time format parser to break up the date format into its parts and implement it in the `DateTimeDragControl`.

### DIFF
--- a/src/LogExpert/Classes/DateTimeParser/Parser.cs
+++ b/src/LogExpert/Classes/DateTimeParser/Parser.cs
@@ -1,0 +1,150 @@
+﻿using System.Collections.Generic;
+
+namespace LogExpert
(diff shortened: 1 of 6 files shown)


igitur:
Let me know if I should explain in more detail what this PR, and the date format parser does.

zarunbal:
Thanks for adapting / updating your PR to the changes. I'll have a look into it soon (Hopefully that I have some more free time). I currently try to finish the Release pipeline so I have less work to release a new version.

zarunbal:
Again thanks for your contribution! It would be great if you could add some unit tests at least for the non gui parts, by time.

igitur:
Sure, I'll add unit tests later.

PR #106, opened 2019-05-17
Unit tests for dateformat parser

igitur:
- Small refactoring
- Fix to support Bulgarian date formats
- Unit tests added


igitur:
I had to exlude the `dz-*` culture names. They don't exist on my Windows 7 machine, so I'm unsure why the date format parser cannot parse them.

zarunbal:
Thank you for the Unit tests and the fix! :-)

PR #108, opened 2019-05-21
Translation of the german comments

Hirogen:
Just translated, the comments, all the assamblyInfo files should now have the english comment.

One real translation in src/LogExpert/Classes/Log/LogfileReader.cs


zarunbal:
Thank you for your translation. The translations are good

Issue #115, opened 2019-06-11
Feature Request: Add a Setup / Installer


Hirogen:
Maybe we should add an Setup / Installer to the project?

This could help with unblocking the dlls/exe, due to the installing process running with certain privileges

Like Inno-Setup or something like that? Inno-Setup itself is open source.

Hirogen:
https://github.com/jrsoftware/issrc

Hirogen:
Currently in process of creating the first inno setup script
here a few screenshots of the work in progress

![image](https://user-images.githubusercontent.com/8225398/59934249-6d605a80-944b-11e9-91b4-d1d9ba3e55ac.png)

![image](https://user-images.githubusercontent.com/8225398/59934261-73eed200-944b-11e9-8bf1-c4a3617673aa.png)

![image](https://user-images.githubusercontent.com/8225398/59934273-78b38600-944b-11e9-9dfc-143a847efd1d.png)

![image](https://user-images.githubusercontent.com/8225398/59934296-8406b180-944b-11e9-9749-86d1220ecced.png)

![image](https://user-images.githubusercontent.com/8225398/59934305-88cb6580-944b-11e9-830e-c831c8e94396.png)

![image](https://user-images.githubusercontent.com/8225398/59934317-8d901980-944b-11e9-879b-d35fd67425c8.png)

![image](https://user-images.githubusercontent.com/8225398/59934323-91bc3700-944b-11e9-995c-a40535392b9e.png)

![image](https://user-images.githubusercontent.com/8225398/59934336-954fbe00-944b-11e9-91a4-b75814de880b.png)

Hirogen:
![image](https://user-images.githubusercontent.com/8225398/59934420-c5975c80-944b-11e9-84e0-1401c300d6e0.png)

![image](https://user-images.githubusercontent.com/8225398/59934430-caf4a700-944b-11e9-8446-974442ee5e31.png)

Hirogen:
Pull Request added #117

Hirogen:
Pull Request has been merged, closing issue

PR #117, opened 2019-06-21
Inno Setup Script and Changelog added

Hirogen:
I added a Inno Setup Script to create an installer.
Currently the installer only checks for .net Framework Installation 4.0+ (can be changed later, to 4.7.2)
Everything else should be self explanatory.
The script uses relative paths, so it should work with all forks

Supported languages for the inno script installer German and English, more can be added

this is for Issue: #115


Hirogen:
Uh, yeah and I also added a "template" for the change log

zarunbal:
Wow Thank you for the inno script! Back from vacation and cool pull requests, this is great to come back :)
Some questions / remarks from my side
- What software / tools are necessary to create the setup?
- Which command line and parameter are needed to create the setup (I would like to include it into the build project)
- Is it necessary to include the Changelog and info page into setup? 
- Is it possible to change the folder to e.g. \bin\SetupFiles? I would currently prefer to create a independent folder with only the needed files and then create from this folder the setup?
- The folder pluginsx86 is unnecessary. This is only for the ssh plugin 
- Exclude the *.pdb files. They are not needed for execution and bloat the setup

Background Infos: Currently I create in build copies for e.g. the chocolatey and other packages with specific subfolder which are only for one package. These subfolders contain only the needed files for the one package.

My opinion for the extra pages, I'm not sure if they are really needed because people have already downloaded the setup from here (maybe other sources) and read the Infos and changelog. Why display them again and force the user to "ignore" more setup pages as needed. 

What are you thinking about a msi? I think this is really easy to automatically install it with different software distribution systems

Hirogen:
> Wow Thank you for the inno script! Back from vacation and cool pull requests, this is great to come back :)

you're very welcome and welcome back, hope you are refreshed :D

> What are you thinking about a msi? I think this is really easy to automatically install it with different software distribution systems

I don't have any experience with MSI, we don't use it where I work, I really haven't tried it.

> * What software / tools are necessary to create the setup?
Only Inno Setup installation => http://www.jrsoftware.org/isdl.php or if you want to build it your self, you can also download the source code => https://github.com/jrsoftware/issrc

Atom, has an Inno Setup compiler! => https://atom.io/packages/build-innosetup
AppVeyor supports it as well => https://www.appveyor.com/docs/windows-images-software/

> * Which command line and parameter are needed to create the setup (I would like to include it into the build project)

compil32 /cc "c:\isetup\samples\my script.iss"
The Setup Compiler will return an exit code of 0 if the compile was successful, 1 if the command line parameters were invalid, or 2 if the compile failed.

> * Is it necessary to include the Changelog and info page into setup?
> My opinion for the extra pages, I'm not sure if they are really needed because people have already downloaded the setup from here (maybe other sources) and read the Infos and changelog. Why display them again and force the user to "ignore" more setup pages as needed.

No really, I just added it for information purpose
I commented them out
;InfoBeforeFile=..\CHANGELOG.md
;InfoAfterFile=..\README.md

At my workplace we add the changelog, so people see what has been different since the last version, and we also add an readme, just to be sure.

> * Is it possible to change the folder to e.g. \bin\SetupFiles? I would currently prefer to create a independent folder with only the needed files and then create from this folder the setup?

Yes, I changed it :)

> * The folder pluginsx86 is unnecessary. This is only for the ssh plugin

done
> * Exclude the *.pdb files. They are not needed for execution and bloat the setup

done

Hirogen:
Today in the evening (CET) I'll have a look at MSI, maybe I can cook up a project

zarunbal:
You can have a look into https://wixtoolset.org/ with this I have some time ago created msi setups

zarunbal:
> >Wow Thank you for the inno script! Back from vacation and cool pull requests, this is great to come back :)

>you're very welcome and welcome back, hope you are refreshed :D

Thank you, it was a short vacation but refreshing 

> At my workplace we add the changelog, so people see what has been different since the last version, and we also add an readme, just to be sure.

In my opinion the people already able read this here on github

You can already work with .NET 4.7.2 as in the current release note stated this will be the next .NET version used (see 20a5bed292ebaebc36d4f57feb110cc93f667299)

Hirogen:
> You can already work with .NET 4.7.2 as in the current release note stated this will be the next .NET version used

done

PR #126, opened 2019-08-11
Missing CSV Columnizer #8

miltob:
Fixed the problem by reverted change around the default columnizer.
Changed the auto search columnizer feature to only search for columnizer when the log file is opened the first time in the LogExpert.
Added a test case.
Updated csv default separator.


src/CsvColumnizer/CsvColumnizer.cs
```diff
@@ -47,7 +47,7 @@ public class CsvColumnizerConfig
 
         public void InitDefaults()
         {
-            delimiterChar = ';';
```
miltob: I am not sure why this was semicolon instead of comma.

miltob: Please let me know if I missed anything.

zarunbal: Depends on the CSV file. There are some which are delimited by semicolon instead of comma. In my knowledge some people use the semicolon because its not used in "sentences" as for the comma its used often and needs some quotation.

PR #135, opened 2019-10-19
Set columnizer to "Default (single line)" by default? #128

### LINKED ISSUE #128: Set columnizer to "Default (single line)" by default?
How do you set the default columnizer to "Default (single line)" ?

Important questions
- LogExpert Version
1.8.0
- Windows Version
Windows 10 x64 1903
- Lokal file or Network
Network
- Used Columnizer
I'm attempting to set default columnizer to: Default (single line). It opens up some particular log files with Timestamp Columnizer though. Because some of the lines in these particular log files have a space before [DATE TIME], it doesn't columnize those lines correctly.  I'm just trying to default LogExpert to "Default (single line)" for now.
- Is this version specific or a new problem
I assume this is a new issue or maybe by-design behavior.  I have older 1.6.13 on a Win7 computer with what looks like identical settings and it defaults to "Default (single line)".  I'm trying to do the same with latest 1.8.0.

Optional questions(depends on the problem):
- Continous appending log file?
Tailing is fine.
- Size of the file
~100kb
- some specific settings 
Settings | Columnizers is empty, but also tried a file name mask with both blank and * with columnizer set to Default w/o luck.
- Screenshot

# Feature
One "Issue" per feature! If in doubt create a new one.

Please provide a rough use case so its possible to get a deeper understanding how the feature would be used.

# General
Please don't forget, this repository is maintened in free time! The tasks will be tackled as soon as there is time.
No problem! Thanks!
cheers,
Scott

miltob:
Give any existing Columnizer matching priority.
This does not fix the first issue in #128 
But this fix the Columnizer RegEx not working issue mentioned in #128 .
To fix #128, I am planning to create a config to allow user to disable auto columnizer.

### DIFF
--- a/src/LogExpert/Controls/LogWindow/LogWindowsPublic.cs
+++ b/src/LogExpert/Controls/LogWindow/LogWindowsPublic.cs
@@ -40,18 +40,17 @@ namespace LogExpert
                 }
 
                 //
-                // beenLoadedBefore is used for automatically find the best columnizer.
-                // When a new log file is opened, this feature will try to find a columnizer instead of the
-                // default one.
+                // isUsingDefaultColumnizer is to enable automatically find the best columnizer.
+                // When a new log file is opened, and no Columnizer can be chose by file mask,
+                // this flag will enable find a columnizer automatically.
                 // Current solution is not elegant.
                 // Since the refactory will involving a lot of work, we can plan it in the future.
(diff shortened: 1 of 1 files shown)


miltob:
@zarunbal I guess something went wrong with the CI pipeline. I am not very familiar with the setup here. Any idea what's problem?

PR #146, opened 2019-12-03
added truncate file option to context menu

jsuppe:
I had found myself deleting the log file(s) repeatedly to get clean view for certain log files, especially when filters would be capturing too much. This option to truncate that log file improved my workflow.


src/LogExpert/Controls/LogTabWindow/LogTabWindowEventHandlers.cs
```diff
@@ -919,6 +920,12 @@ private void findInExplorerToolStripMenuItem_Click(object sender, EventArgs e)
             explorer.Start();
         }
 
+        private void truncateFileToolStripMenuItem_Click(object sender, EventArgs e)
+        {
+            LogWindow logWindow = dockPanel.ActiveContent as LogWindow;
+            File.WriteAllText(logWindow.Title,"");
+        }
+
```
Hirogen: Does this take into account, if the file is not write able by another process?
This is not a safe way to create a new file => Exceptions are not taken into account!

zarunbal:
The easier way would be in my opinion to use the "external tools" feature. There you can launch any executable or e.g. cmd.exe with some commands.

den-rain:
I can vote for the feature to truncate a log file! It will be very helpful.

jsuppe:
New update to this branch checks for the file lock and writes to status what the process holding the lock is named. Also, this is now described as "try to" truncate, because it is not guaranteed. This change only aims to address when exception may be thrown due to trying to exclusively obtain write lock on file that already has lock and would then cause an exception.

Hirogen:
please merge the latest changes from the development branch, you got over 200 files, that would change back to an older version if I accepted that MR

Hirogen (changes requested):
sorry for the late answer, could you please add unit testcases, otherwise it looks ok

Neustradamus:
@jsuppe: Can you update your draft PR?

Issue #159, opened 2020-04-29
http://www.log-expert.de/ no longer related?


ansorg:
hi

just installed 1.8.5 and noticed that the "About" screen still points to http://www.log-expert.de/ which has some strange content (as of today) 

![grafik](https://user-images.githubusercontent.com/3541784/80601688-6b353800-8a2e-11ea-953e-e4af0cf54d73.png)

solarpolar:
It means "go away" :(

Hirogen:
see pull request #160 , changed it to the new github url

zarunbal:
See new release, I'll close this issue. If there is still a problem feel free to reopen this issue :)

PR #160, opened 2020-04-30
changed url in the about window

Hirogen:
removed old .de url and added the new github url

#159


zarunbal:
This is already done in the pull #158 if I see right? Then dis can be closed without merge?

Hirogen:
> This is already done in the pull #158 if I see right? Then dis can be closed without merge?

yes, seems so :)

PR #168, opened 2020-05-11
Column context menu was not working

joostaarts:
I was unable to hide a column because the context menu for the gridview header seems to be overridden by the context menu for the grid cells. Please have a look at my PR.


zarunbal:
Thank you for your pull request with the bug fix :)

Issue #191, opened 2021-02-25
Multi-file doesn't guarantee file ordering


natenho:
# Bug

Important questions
- LogExpert 1.70 (also reproducible in last version)
- Windows Server 2012 R2 (also reproducible in Windows 10)
- Local
- Used Columnizer: No

According to https://stackoverflow.com/questions/38319915/c-sharp-openfiledialog-to-select-multiple-files-with-same-order-as-user-selects/38319967#comment117205495_38319967, the order of the selected files are not guaranteed in multi-file.

Such behavior causes log files to be shown in undesirable order.

![image](https://user-images.githubusercontent.com/4236481/109083171-288ecb80-76e4-11eb-8286-0a77a4054d68.png)

Hirogen:
#192 merged and added this feature

PR #192, opened 2021-02-25
Fix file order selection when using multi-file because it is not guaranteed by OS

natenho:
@zarunbal 

According to https://stackoverflow.com/a/38319967/1987788, it is not possible to get the actual selection order, so it was causing wrong log ordering in some cases.

The implementation simply sort the selected files by name.


natenho:
This pull request closes #191

Hirogen (approved):

PR #193, opened 2021-02-25
Allow resize and add autosize to the filter box (close #177)

### LINKED ISSUE #177: [Feature request] Resizable "Text filter" field
# Feature

image

Can we increase width of "Text filter" field, or could that be automatically resized according to number of characters present?
I am using Regular expression and more than one command to filter out the result, but this field is quite short.
Also history of text filter is with fixed width and if there ase similar commands, they all looks same.

natenho:
@zarunbal please review the code. You'll notice some misplaced code in form designer code, unfortunately VS2019 seems to be not too friendly with multiple partial form files. IMHO all these logic should be moved to specialized classes instead of partial form files.

- Dragging the splitter allow free filter box resizing
- You can see that a double click will auto resize the filter box expanding to the maximum text width
- The filter box font was changed so it can be easier to read and relate to the log itself

logexpert

### DIFF
--- a/src/LogExpert/Controls/LogWindow/LogWindow.cs
+++ b/src/LogExpert/Controls/LogWindow/LogWindow.cs
@@ -173,8 +173,9 @@ namespace LogExpert
                 filterComboBox.Items.Add(item);
             }
 
-            filterComboBox.DropDownHeight = filterComboBox.ItemHeight * ConfigManager.Settings.preferences.maximumFilterEntriesDisplayed;
-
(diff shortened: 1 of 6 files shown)


Hirogen (commented):
In filterSplitContainer_MouseMove please cast the object only once
ex: SplitContainer container = (SplitContainer)sender;

this will make the source code clearer to read and also it is only converted once, not multiple times

other than that looks fine for me

PR #203, opened 2021-10-11
Prevent ArgumentOutOfRangeException when navigating through Multifiles (close #202)

### LINKED ISSUE #202: Navigating through MultiFiles
When Navigating through MultiFiles using alt-up/alt-down there is an error when the lastmost boundary is reached. But only if the newest logfile is empty (no lines produced yet).
I guess the alt-down in MultiFile-view is trying to jump to the first line in that newst file, but since it is empty there is an _ArgumentOutOfRangeException_ exception. Verified by adding a line in the emptty file and the error went away.

LogExpert 1.8.7 on Win10 with local logfiles (copied unmodified from apache2-logs on a raspberry pi)

See stacktrace below

```
Index låg utanför intervallet. Det får inte vara negativt och måste vara mindre än mängdens storlek.
Parameternamn: index

ArgumentOutOfRangeException
   vid System.Collections.ArrayList.get_Item(Int32 index)
   vid System.Windows.Forms.DataGridViewRowCollection.SharedRow(Int32 rowIndex)
   vid System.Windows.Forms.DataGridViewRowCollection.get_Item(Int32 index)
   vid LogExpert.LogWindow.SelectLine(Int32 line, Boolean triggerSyncCall, Boolean shouldScroll)
   vid LogExpert.LogWindow.LogWindow_KeyDown(Object sender, KeyEventArgs e)
   vid LogExpert.LogTabWindow.LogTabWindow_KeyDown(Object sender, KeyEventArgs e)
   vid System.Windows.Forms.Control.OnKeyDown(KeyEventArgs e)
   vid System.Windows.Forms.Control.ProcessKeyEventArgs(Message& m)
   vid System.Windows.Forms.Form.ProcessKeyPreview(Message& m)
   vid System.Windows.Forms.Form.ProcessKeyPreview(Message& m)
   vid System.Windows.Forms.Control.ProcessKeyMessage(Message& m)
   vid System.Windows.Forms.Control.WmKeyChar(Message& m)
   vid System.Windows.Forms.Control.WndProc(Message& m)
   vid System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

hakanostrom:
Preventing `ArgumentOutOfRangeException` when navigating through Multifiles, caused in the case when the newest file "in line" is empty (no loglines in it).

### DIFF
--- a/src/LogExpert/Controls/LogWindow/LogWindowPrivate.cs
+++ b/src/LogExpert/Controls/LogWindow/LogWindowPrivate.cs
@@ -1706,6 +1706,12 @@ namespace LogExpert
                     return;
                 }
 
+                // Prevent ArgumentOutOfRangeException
+                else if (line >= dataGridView.Rows.Count)
+                {
+                    line = dataGridView.Rows.Count - 1;
+                }
+
                 dataGridView.Rows[line].Selected = true;
 
                 if (shouldScroll)
@@ -1713,6 +1719,10 @@ namespace LogExpert
                     dataGridView.CurrentCell = dataGridView.Rows[line].Cells[0];
                     dataGridView.Focus();
                 }
+            }
+            catch (ArgumentOutOfRangeException e)
+            {
+                _logger.Error(e, "Error while selecting line: ");
             }
             catch (IndexOutOfRangeException e)
             {


Hirogen (commented):
this looks good, but quick question, the else if does prevent the ArgumentOutOfRangeException, but it is still catched, did you have cases where the ArgumentOutOfRangeException still happend, even with the else if?

hakanostrom:
It was made as a double safety. The elseIf adresses the bug/issue and prevents ArgumentOutOfRangeException.
The catch is there only for other possible cases where ArgumentOutOfRangeException might be thrown (but in this case, none I can think of).

PR #207, opened 2021-11-28
merge development to master

Hirogen:
merge development into master and also:
Added parsesing support for old latin date format "gg"
Latin Date String: (a.C.n.  ante Christum natum)


Hirogen:
preparation for release 1.8.8 if i get the versioning to work

PR #216, opened 2022-03-27
updated the nuget packages to use newer versions

Hirogen:
updated the nuget packages to use newer versions

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,0 +1,5 @@
+mode: ContinuousDelivery
+branches: {}
+ignore:
+  sha: []
+merge-message-formats: {}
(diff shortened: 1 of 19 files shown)


Hirogen:
somehow gitversion works on my machine but not with the appveyor ci -.-

Hirogen:
something with the GitVersion.BranchName or the GitVersion.SHA...

Hirogen:
GitVersion.BranchName and GitVersion.SHA currently not working, with GitVersion 5.9.0, probably this problem => https://github.com/GitTools/GitVersion/pull/3033
Commted it out, so it is not used

PR #223, opened 2022-06-20
Json options read and write

Hirogen:
Settings will now be read and written as JSON File!
If somebody need to create a json file, ReRelease Version 1.8.7 can save the settings as json


Hirogen:
This will fix: !219

Petro31:
> Settings will now be read and written as JSON File! If somebody need to create a json file, ReRelease Version 1.8.7 can save the settings as json

Uh, how?  There's no option for json export of the settings...

Hirogen:
the rerelease of version 1.8.7 saves the settings as json file as well when you press save

PR #225, opened 2022-06-21
Highlight import no json selectable

Hirogen:
Highlight import no json was selectable, still .dat file


Hirogen:
Also renamed Events to use OnEvent

Issue #182, opened 2020-08-12
Fonts and UI is pixelated on High DPI screens


rkriauciukas-freesat:
Windows 10, using the latest version of LogExpert@1.8.7. Surface Book 2, so laptop with DPI of 200% in Windows settings.

The UI of LogExpert is quite pixelated, including fonts. None of the option combinations in "Change high DPI settings" via Properties of logexpert.exe fix it. Some options fix the fonts, but then Menubar is under Toolbar etc.

Anyone seeing the same?

Hirogen:
As far as I know, Logexpert does not support Higher DPI than 100%

rkriauciukas-freesat:
Ah, I wish it did! It looks to be a great tool, hopefully it can be updated soon :)
Thanks!

Hirogen:
> Ah, I wish it did! It looks to be a great tool, hopefully it can be updated soon :)
> Thanks!

Honestly, never looked at this kind of stuff, so don't know how much "work" it could be. Doesn't look like a small change :D

rkriauciukas-freesat:
Most of the High DPI scaling issues on the apps can be fixed with a right combo in the *.exe Properties -> High DPI Settings. But I could not find a combination which would make LogExpert look properly: fonts get fixed and look nice, but then menu icons are super tiny and they are above menu bar. 

So maybe there is no need for .net 4.7, but just to find a way to make it work with one of those compatibility settings:
![image](https://user-images.githubusercontent.com/64024442/90111870-e8738700-dd57-11ea-8b67-a67569b0becb.png)

Hirogen:
> Most of the High DPI scaling issues on the apps can be fixed with a right combo in the *.exe Properties -> High DPI Settings.

that's awesome, but this can not be the real solution, only temporary fix. If High DPI will be supported it should be properly implemented

Hirogen:
https://docs.microsoft.com/en-us/dotnet/desktop/winforms/high-dpi-support-in-windows-forms?view=netframeworkdesktop-4.8

there is a possibility to add this, but it needs a few tweeks as can be seen in this screenshot (4k Display)
<img width="674" alt="image" src="https://user-images.githubusercontent.com/8225398/179362248-75fdbf49-00c9-4633-ab26-f89f6d63d10e.png">

Hirogen:
[...]

PR #238, opened 2022-07-16
Issues/issue 182 high dpi

Hirogen:
this a draft, there is still work to be done, all forms have to be updated

#182


Hirogen:
Known Problems:
TabHeader Texts are very small:
image

Textfilter (F4) and Pro Edition are cropped and not displayed correctly
image

PR #251, opened 2022-07-26
Dark Mode v1

jrouzies-fr:
Implement Dark Mode option:

- Added in Settings
- Converts everything to Dark Mode excepts:
  - Tabs line 
  - Dialog box menus (like Settings).
  - Default log background. Workaround: apply an highlight for everything to be darker.


jrouzies-fr:
Some preview:

image

Menu:

image

Setting:

image

Hirogen:
- Please extract the classes. 1 Class - 1 File. (When I refactor something I also extract the older classes from the files)
- Put the Form Extensions under the following folder:
  - Logexpert / Extensions / Forms
- please use paletton.com or any other tool to get a "dark color scheme", and change the colors accordingly example: https://paletton.com/#uid=75-0u0k00bN0anK00hJ014l6n0u
- if possible avoid black at all costs, use darker greys, best examble is the visual studio dark theme!
To change the color of the Grid, you need to adapt the class BufferedDataGridView

jrouzies-fr:
> * Please extract the classes. 1 Class - 1 File. (When I refactor something I also extract the older classes from the files)
> * Put the Form Extensions under the following folder:
>   
>   * Logexpert / Extensions / Forms
> * please use paletton.com or any other tool to get a "dark color scheme", and change the colors accordingly example: https://paletton.com/#uid=75-0u0k00bN0anK00hJ014l6n0u
> * if possible avoid black at all costs, use darker greys, best examble is the visual studio dark theme!
> * To change the color of the Grid, you need to adapt the class BufferedDataGridView
> * Please us the "Dark-Theme" branch, I just created it, it will server as the basis for all changes the still need to be done

I implemented most, but I am blocking on those 2 points:

1. I can't find where are the Log Tabs visible in the code and where to change their colors
2. The ForeColor on the dataGridView cells don't seem to work, I think it might apply the Tab foreColor

image

Hirogen:
>I can't find where are the Log Tabs visible in the code and where to change their colors

that would be part of the WinFormsUI - The "DockPanel" if I'm not mistaken!

> The ForeColor on the dataGridView cells don't seem to work, I think it might apply the Tab foreColor

that would be the function

`public void CellPainting(DataGridView gridView, int rowIndex, DataGridViewCellPaintingEventArgs e)`

in the LogWindowsPublic.cs file

jrouzies-fr:
> 

I was already modifying in the CellPainting, the background is working perfectly, but not the ForeGround

image

I'll continue checking tomorrow!

jrouzies-fr:
> > 
> 
> I was already modifying in the CellPainting, the background is working perfectly, but not the ForeGround
> 
> image
> 
> I'll continue checking tomorrow!

The text color was later, this is fixed.

-------

I am still struggling for the Tabs, they seem to be inserted programmatically but there is no TabControl with its colors.

There was a commented method here (LogTabWindowPrivate), I suppose this is why changing the color tab from the menu doesn't work:

image

I found the code to change the Text tab but that's it, no colors options.

-------

Also I don't seem to find a solution to stretch the Timeshift textbox all the way, from what I searched not sure it is doable:

image

jrouzies-fr:
Ok it seems to be all those gradient stuff

image

image

Hirogen:
> There was a commented method here (LogTabWindowPrivate), I suppose this is why changing the color tab from the menu doesn't work: 

menu does look way  better then the pure black one, color scheme is really nice, the red though is kind of eye destroying ;D

jrouzies-fr:
> > There was a commented method here (LogTabWindowPrivate), I suppose this is why changing the color tab from the menu doesn't work:
> 
> menu does look way better then the pure black one, color scheme is really nice, the red though is kind of eye destroying ;D

The red is just to make sure highlights are still working through code changing :P

I was able to make the tabs work !

image

jrouzies-fr:
I renamed the branch I will re-do a PR

jrouzies-fr:
@Hirogen  on which branch should I merge the Dark-Theme? There is already one on your side no?

Hirogen:
yes, dark-theme is the designated branch, until we could verify it works as it should be, looks good and has no further impact on the rest

PR #257, opened 2022-08-05
Choco package enhancement

Matze693:
I added two feature to the chocolatey package:

1. Shortcut in start menu
image

2. Context menu (right-click) entry to open a file directly with LogExpert
image

Both feature are only applied for current windows user.


Hirogen:
will review tomorrow

Hirogen (approved):

PR #260, opened 2022-08-18
Added high DPI awareness

aseemsahoo:
Added high DPI awareness, enabling per-monitor DPI awareness in the app.config file.


Hirogen:
Sadly, app.config settings are not enough, but there is already a branch for that => https://github.com/zarunbal/LogExpert/tree/High-DPI-Test and already a test-version => https://github.com/zarunbal/LogExpert/releases/tag/v1.9.0.High.DPI

Hirogen:
See also the conversation here: https://github.com/zarunbal/LogExpert/issues/182

PR #262, opened 2022-08-19
Issues/issue 245 multi instance warning

Hirogen:
First Draft, to let user no longer show the error message


Hirogen:
If one Instance is open and the error message should be blocked, the open instance settings dialog is not updated, because it does not load the settings on opening the settings dialog (this needs to be added)

PR #276, opened 2023-03-08
Added code to the RegexColumnizer to avoid null columns & the resulting NullPointer exceptions

kniveg:
I have added some code to the RegexColumnizer that will add columns with empty string when the line doesn't match the regex. 

The reason I made this alteration is to fix the following problem I have been having: If you use this RegexColumnizer to read logs that contain stack traces (lines that don't match the regex)  you get null columns for that line. This is visualized ok but will result in many error message pop-ups when searching in these columns making it hard to use this columnizer for these kind of logs. 

The added code aims to fix this problem by eliminating the null columns. This doesn't change the visualisation but fixed the search problem.

### DIFF
--- a/src/RegexColumnizer.UnitTests/RegexColumnizerTests.cs
+++ b/src/RegexColumnizer.UnitTests/RegexColumnizerTests.cs
@@ -1,0 +1,79 @@
+﻿using LogExpert;
+using Moq;
+using NUnit.Framework;
+
+namespace RegexColumnizer.UnitTests
+{
+    [TestFixture]
+    public class RegexColumnizerTests
+    {
+        [Test]
+        public void SplitLineLogRegexDoesNotMatchLine()
+        {
+            var columnizerConfig = new RegexColumnizerConfig
+            {
+                //We just need a regex that matches the line.
+                Expression = @"^(?'time'[\d]+)\s+(?'Message'.+)$",
+                Name = "Test regex"
+            };
+
+            var columnizer = new Regex1Columnizer();
+            columnizer.Init(columnizerConfig);
+            
+            var notMatchingLogLine = new TestLogLine(5, "Error in com.example.core");
+            
+            var parsedLogLine = columnizer.SplitLine(Mock.Of<ILogLineColumnizerCallback>(), notMatchingLogLine);
+            
+            // verify that the expected number of columns are created
+            Assert.AreEqual(2, parsedLogLine.ColumnValues.Length);
+
+            //verify that the entire text is in the last column
+            Assert.AreEqual("Error in com.example.core", parsedLogLine.ColumnValues[1].Text);
+            
+            // Verify that the other columns are empty strings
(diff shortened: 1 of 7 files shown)


src/RegexColumnizer/RegexColumnizer.cs
```diff
@@ -71,6 +71,16 @@ public IColumnizedLogLine SplitLine(ILogLineColumnizerCallback callback, ILogLin
...
                         FullValue = line.FullLine
                     };
+                    
+                    //Fill other columns with empty string to avoid null pointer exceptions in unexpected places
+                    for (int i = columns.Length - 2; i >= 0; i--)
+                    {
+                        logLine.ColumnValues[i] = new Column
+                        {
+                            Parent = logLine,
+                            FullValue = ""
```
Hirogen, line 81: always use "string.Empty"

Hirogen (changes requested):
reviewed please adapt

src/RegexColumnizer/RegexColumnizer.cs
```diff
@@ -71,6 +71,16 @@ public IColumnizedLogLine SplitLine(ILogLineColumnizerCallback callback, ILogLin
                         Parent = logLine,
                         FullValue = line.FullLine
                     };
+                    
+                    //Fill other columns with empty string to avoid null pointer exceptions in unexpected places
+                    for (int i = columns.Length - 2; i >= 0; i--)
```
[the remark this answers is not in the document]
kniveg: The loop is reversed because the other loops in this file are also reversed and I was trying to use the same code style as previously used. That is also where the -2 comes from. I updated the loop to be in assenting order. 

A foreach loop is not possible here (IMHO) since the array is modified in the loop.

kniveg:
I made the requested changed. my apologies for the extra commits. I had some trouble getting the build to go green (C# is not the language I usually work in).

src/RegexColumnizer.UnitTests/RegexColumnizerTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            };
+
+            var columnizer = new Regex1Columnizer();
+            columnizer.Init(columnizerConfig);
+            
+            var notMatchingLogLine = new TestLogLine(5, "Error in com.example.core");
+            
+            var parsedLogLine = columnizer.SplitLine(Mock.Of<ILogLineColumnizerCallback>(), notMatchingLogLine);
+            
+            // verify that the expected number of columns are created
```
Hirogen, line 27: Every Assert should be its own testcase, if those are just different data-sets, add the data to the function header (see the examples in "LogExpert.Tests" project :)

Hirogen (changes requested):

src/RegexColumnizer.UnitTests/RegexColumnizerTests.cs
Hirogen: thanks for contributing 👍

Hirogen (approved):

PR #277, opened 2023-03-09
copy to clipboard: always add line number

### LINKED ISSUE #275: copy to clipboard: always add line number
# Feature
when copying a line from the "filtered lines", the line number is prefixed.

when copying from the main view, the line numbers are not added.   It would be nice if we can have the same format for pasted lines from both views.   

It also makes it easier to create a separate file with selected log lines.    I always copy the "needed" lines to a separate file (with different filters) and sort them by line number.

pentop:
fixes #275

### DIFF
--- a/src/ColumnizerLib.UnitTests/ColumnizerLib.UnitTests.csproj
+++ b/src/ColumnizerLib.UnitTests/ColumnizerLib.UnitTests.csproj
@@ -46,6 +46,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="ColumnTests.cs" />
+    <Compile Include="Extensions\LogLineExtensionsTests.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


src/ColumnizerLib.UnitTests/ColumnizerLib.UnitTests.csproj
```diff
@@ -59,10 +60,10 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="NUnit">
-      <Version>3.13.3</Version>
+      <Version>3.12.0</Version>
     </PackageReference>
     <PackageReference Include="NUnit3TestAdapter">
-      <Version>4.2.1</Version>
+      <Version>3.15.1</Version>
```
Hirogen, line 66: Why is the Nunit Packaged downgraded?

Hirogen:
Please change the branch from master to development!

src/ColumnizerLib.UnitTests/ColumnizerLib.UnitTests.csproj
pentop: ooops! Fixed

PR #280, opened 2023-03-23
Replaced chilkat proprietary dll with open source ssh.net

Hirogen:
Replaced the proprietary chilkat dlls with the open source ssh.net (https://github.com/sshnet/SSH.NET)


Hirogen:
Tested with following formats and rolling logfiles

Uri format
sftp://username:password@127.0.1.0/var/log/logfile

IP Adress + Port + File
sftp://127.0.0.1:1234/logfile.log

PR #286, opened 2023-06-18
Making sure the window is located within the current virtual desktop.

SoftAllan:
This fix looks at the stored appBounds information and make sure that the values are within the current virtual desktop bounds. It can happen that when you work on different virtual desktops the stored information is out of bounds making it impossible to locate or restore the view position. This is now checked and if the values are out of bounds the default windows bound setting is used.


Hirogen:
please change the branch to development

Hirogen:
We have double the Source-Code, this is not good, you could instead change it so, that the calculation of is done within the ConfigManager. This would avoid 2 Times the same function in different places.

SoftAllan:
Yes I was thinking about that too. I'll have a look at that.
Should I close this PR and create a new when I have something new to push? I haven't used Github as much for Pull Request. I'm more familiar with Microsoft DevOps.

SoftAllan:
Hmm... rebasing is not working good for github. I'll make the new changes on a new branch from Development.

PR #293, opened 2024-02-28
Update Versions, Framework and NuGet Packages

KaiL3:
Update Framework to 4.8.1
Update .netcore 3.1 => .net 8
Update all NuGet Packages
Depend on Update the TestPackage, migrate the Testcode AreEqual => That()

### DIFF
--- a/.nuke
+++ b/.nuke
@@ -1,1 +1,0 @@
-src/LogExpert.sln
--- a/.nuke/build.schema.json
+++ b/.nuke/build.schema.json
@@ -1,0 +1,190 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "$ref": "#/definitions/build",
+  "title": "Build Schema",
+  "definitions": {
+    "build": {
+      "type": "object",
+      "properties": {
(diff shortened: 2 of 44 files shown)


KaiL3:
Sorry for this confusion, but i would like opening a new Pullrequest from the Development Branch to the Development Branch. In this request are included the .net 8 update from the build project.

KaiL3:
Hi Patrick, I'm surprised. When testing this pull request, it was compiled against .NET 4.7.2. Actually I would have expected .NET 4.8.1 and after the merge the developer pack for the new .NET version was missing. 
Have you also noticed this?

Hirogen:
not really, since target framework is 4.8.1, very strange

KaiL3:
The Developer Pack .NET 4.8.1 must be installed in the pipeline for compilation. Alternatively, a downgrade to .NET Framework 4.8 should also work.
It is interesting that the initial tests did not run on 4.8.1, but on 4.7.2. I tested the build locally, where everything worked. Except for the setup and the package manager, which I did not install locally.

PR #300, opened 2024-12-30
.Net 8 - Build script fix

RandallFlagg:

PR #301, opened 2025-01-06
High DPI for .Net 8 

TheNicker:
Add support for high DPI monitors.
Fixes the blurry text in the program
Depends on #300


Hirogen (approved):
Thanks for your contribution!

PR #303, opened 2025-01-29
Update LogStreamReaderBase.cs

CzmQlik:
Avoids NullReference on Finalizer on GC thread. A finalizer should always call Dispose with disposing set to false because some objects could already have been collected and set to null.

Dispose() (and so, with disposing set to true) is expected to be manually called in .net (all objects that implement IDisposable should be used together with `using` or Dispose(true) must be manually called by the developer, if it's feasible).
In any case, a finalizer is meant just as last resort (and in fact, in the general case, it is suggested not to implement it).

See https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/dispose-pattern#finalizable-types and the whole dispose-pattern page.

**Exception:**

_Object reference not set to an instance of an object.
NullReferenceException
   in LogExpert.Classes.Log.PositionAwareStreamReaderBase.Dispose(Boolean disposing)
   in LogExpert.Classes.Log.LogStreamReaderBase.Finalize()_

immagine

### DIFF
--- a/src/LogExpert/Classes/Log/LogStreamReaderBase.cs
+++ b/src/LogExpert/Classes/Log/LogStreamReaderBase.cs
@@ -15,7 +15,7 @@ namespace LogExpert.Classes.Log
 
         ~LogStreamReaderBase()
         {
-            Dispose(true);
+            Dispose(false);
         }
 
         #endregion


Hirogen:
please change to [Development](https://github.com/zarunbal/LogExpert/tree/Development) or [.net 8](https://github.com/zarunbal/LogExpert/tree/dot-net-8) branch, better would be .net 8 because I will sooner than later, merge that one back to the dev branch

Neustradamus:
@Hirogen: Can you work on default branch?

It will be better to all.

Currently the project seems dead since several years, latest commit 2022-07-26 and latest version:
- 2022-07-20: https://github.com/zarunbal/LogExpert/releases/tag/v1.9.0

Hirogen:
hi, no, I will not just merge something in the master branch, please change the target to development or .net 8

the project is not dead, because the working branch is development

CzmQlik:
Done https://github.com/zarunbal/LogExpert/pull/310

PR #308, opened 2025-02-03
Dot net 8

Hirogen:


Hirogen:
some UI Changes need to be made, after that, the development branch will be switched to .net 8 and a week after that, the master branch

PR #324, opened 2025-03-02
Issue321: Highlights are now there own .json file

Hirogen:


RandallFlagg:
The reason is because of this commit: [f5e352a7ca3a2739526254f24d6369903c37135c](https://github.com/LogExperts/LogExpert/commit/f5e352a7ca3a2739526254f24d6369903c37135c)

@Hirogen Why did you revert it?

Hirogen:
there are 2 jsons, one is the settings file, with all the settings, the other is the Highlights file, with only the highlights.
The highlights file must be exported / imported from the Highlights-Dialog
image

and that works fine with the dev branch, but maybe I miss understand whats the problem.
image

Hirogen:
but there has been another PR with a breaking change, because the json was not imported / exported correctly
https://github.com/LogExperts/LogExpert/pull/384

TheNicker:
For some reason I was under the impression that highlights.json has been separated from settings.json.
In any way fields like the following should be under different file, e.g. SessionContext / SessionData.
`fileHistoryList`
`lastOpenFilesList`
`searchHistoryList`

My trick is to think about it is: Is this field is redistributable ? can I package and ship it as a predefined settings

Hirogen:
> For some reason I was under the impression that highlights.json has been separated from settings.json. In any way fields like the following should be under different file, e.g. SessionContext / SessionData. `fileHistoryList` `lastOpenFilesList` `searchHistoryList`
> 
> My trick is to think about it is: Is this field is redistributable ? can I package and ship it as a predefined settings

yes but it's a balancing act, I could create 3-4-5 different settings files, and then, I need someway to edit those files and export / import them, also for the user, so many different files are more annoying then a blessing.

Highlights got their own file, because they are separated anyway.

But I agree all the settings should be refactored, since there is also the persister that has it's own settings file

PR #364, opened 2025-03-21
Replaced the WinFormsUI source code with the NuGet version

RandallFlagg:
@Hirogen If there is an option to change the theme, please write me the steps how it is done. If there isn't, you can merge if all is good.


Hirogen (approved):

Hirogen:
> @Hirogen If there is an option to change the theme, please write me the steps how it is done. If there isn't, you can merge if all is good.

there is none, but we can add one later, if needed be, but I don't think we will need one any time soon

Issue #157, opened 2020-04-25
Rollover Detection Not Working?


AvramM:
v1.8.5

Has rollover detection been removed? App no longer seems to detect that the log file has rolled. I have to hit F5 or reload the log file.

Log files are named thus:

MyApp.log
MyApp.log.1
MyApp.log.2 etc

Hirogen:
not that I know of, has the new file the same name?

zarunbal:
so in the previous version e.g. 1.8.2 it worked as expected?

Hirogen:
@AvramM could you try this with 1.8.7 and with 1.8.2 and see if it works?

Hirogen:
is fixed in 1.40

Issue #129, opened 2019-08-29
File automatically tailing


jmhg92:
I often see that when my log file reaches a certain size (aproximately 25MB), it reloads the file and tails it after refreshing, even when "Follow tail" is disabled.

I also noticed that whenever I reach the end of the file the "Follow tail" option is enabled and stays like that until I disable it again.

EDIT:
Whenever I refreshed the file it automatically went to the end of the file and re-enabled tailing.

# Bug
One issue per problem! If in doubt create a new one.
If available attach the generated log file (see wiki)

Important questions
- LogExpert Version
1.8.0

- Windows Version
Windows 7 x64

- Lokal file or Network
Local File

- Used Columnizer
No

- Is this version specific or a new problem
Specific

Optional questions(depends on the problem):
- Continous appending log file?
Yes

- Size of the file
25mb

- some specific settings 
- Screenshot

fschuh:
Also noticed this happening, but on Windows 10 every time the file reloads the UI glitches and gets fully reconstructed which is extremely slow and makes it unusable.
A fix would be appreciated.

rutschi100:
I have the same problem (Windows 10), but to a file that is on the network and to which a lot of stuff is written. I like to work with LogExpert very much but currently I can only use it for files that are rarely written to because of the bug. Use NotePad++ for the other files now (which I don't like that much).

tstreibl:
I have the same problem, using version 1.9.0. 

After unchecking the follow tail option it get's reactivated by itself after a few seconds . This way I can't search something in the log. That's really a problem.

voltz5000:
I have the same problem, using version 1.9.0.

After unchecking the follow tail option it get's reactivated by itself after a few seconds .

RandallFlagg:
@Hirogen Issues #16 and #40 are related to this

Hirogen:
I'll have a look in the evening, I can reproduce #16 and #40 easily on my working machine and if those are gone, I close them

Hirogen:
please check with the new version https://github.com/LogExperts/LogExpert/releases/tag/v.1.12.0

Issue #40, opened 2018-03-14
Frequent black screen when log is updated


natenho:
# Bug

Very often I see the log turned to black after an update. It only redraws the screen after using the scroll bar.

- LogExpert Version 1.6.7
- Windows Version 10
- Local File
- Used Columnizer Default

- Continous appending log file? Yes
- Size of the file ~ 40MB
- Screenshot 
![image](https://user-images.githubusercontent.com/4236481/37384529-43731de2-272e-11e8-9848-537c42ea1513.png)

Pr0metheus2:
The workaround is to tick "Use legacy file reader (slower)" under "Memory/CPU" tab in Options/Settings...

zarunbal:
Thanks for your issue. Do you have this behavior only since in 1.6.7 or is this in old Versions too?

Pr0metheus2:
All versions are affected by this bug...

natenho:
The same issue ocurs in previous versions.
The workaround above didn't work =(.

Titus-s:
Hi, 
I've noticed the similar issue. 
While fast scrolling the log by mouse wheel when LogExpert is not active frequently appears black blocks. 
It seams to be important that LogExpert is not active because when it's active issue doesn't occur. 
![2018 03 29_14h55m15s_014_](https://user-images.githubusercontent.com/11790750/38090310-7e76294e-3362-11e8-8fbd-a4f67d561b2b.png)
Hope it'll help to reproduce this bug.

Hirogen:
Could you test this with the latest version?
I had similar effects in old versions as well, but hadn't any with the latest (1.6.14)

Link to 1.6.14: https://github.com/zarunbal/LogExpert/releases/tag/v1.6.14

natenho:
I've had the same problems using 1.6.14.
Black screens and constantly freezing (screen not redrawing or getting transparent)

![image](https://user-images.githubusercontent.com/4236481/57040181-05ed1200-6c35-11e9-9a5e-c4acb9277a8b.png)

cpatchett-iqbo:
Still an issue in 1.7.1

Hirogen:
is this still an issue with 1.8.7 ?

PiotrPS:
Still present in 1.9.0. Scrolling up and down results in white background.
![image](https://github.com/zarunbal/LogExpert/assets/13746299/efb2bef5-4382-4f7d-996f-e7dfa5b11b36)

[...]

Issue #16, opened 2017-11-27
Sometimes appended file is reloaded completely instead of updated


petsuter:
Usually when a log file is appended to, the only visible update is that the new lines are shown.

But sometimes (almost always with larger files?) the entire file seems to get reloaded (with a progress bar in the statusbar) and the line/text columns are removed and re-added. This is hugely distracting and slow and makes the program useless for such logfiles where this happens constantly.

Hirogen:
This can be tested with continous written loglines (50-100+) per second in very large-Logfiles (800+ mb), that are opend via a network drive.

zarunbal:
Thanks I'll try to reproduce this problem. 

@Hirogen does this behavior only happen after 800+ MB files? How big (characters and Encoding) are your lines (avg, just a rough estimation). Have you some special settings (highlighting, Actions, Buffer Sizes, ...)

Pipecity:
I'm also seeing this behavior but on smaller files as well. I'm using default settings. 
Really frustrating.

isosphere:
I've had this issue, or perhaps a different issue: A small file (~23 MB) would take awhile to load, then upon loading it would be reloaded from scratch. The result is that the file is just continuously reloading.

I've solved the problem in my particular case by selecting "Use legacy file reader" under "Memory/CPU".

nicolas-raoul:
Same symptoms (text and "Line"/"Text" coluns reloading constantly) with a 200KB (yes very small) file updated every 5 seconds on Windows 10 Pro 1803 / LogExpert 1.6.13.

The workaround suggested at https://github.com/zarunbal/LogExpert/issues/16#issuecomment-428218551 solved the problem (now only the few new lines appear and the existing content moves smoothly).

Hirogen:
> 
> Same symptoms (text and "Line"/"Text" coluns reloading constantly) with a 200KB (yes very small) file updated every 5 seconds on Windows 10 Pro 1803 / freshly installed LogExpert 1.6.13 with default settings.
> 

Can you reproduce the problem with the 200kb file all the time it is loaded?

nicolas-raoul:
@Hirogen Yes I just switched `"Use legacy file reader"` off, closed LogExpert, reopened it, and the problem reproduced. Closed and reopened again, it reproduced again. The file is still under 1MB.

Pr0metheus2:
[...]

PR #366, opened 2025-03-22
Removed the thread that watches file changes

RandallFlagg:
@Hirogen can you test this and tell me if there are any issues? I don't know all the places this can be reached by and also can't test all of them as most of the time I am on Linux.

Should improve system resources and handle file event much better.

Might affect the following issues: #16, #40, #129, #157.


Hirogen:
i'll check, first thing that comes to my mind, is a logfile that is loaded from a network drive and updated on said network drive.

I'll test that on Monday, since I don't have that kind of setup at home

Hirogen:
I'll tested it a little bit

this is version 1.12 version and your branch both have the same result
10 ms writes are impossible to handle currently, logexpert just freezes.
20 lines per 50 ms are impossible to handle currently, logexpert just freezes.
10 lines per 50 ms can be handled but i slows down considerably, but still you can use the ui.
<10 lines per 50ms easily handled.

I tested it also with the version 1.9.0 and
10 ms writes are easily handled, no slowdown, regardless of how many lines (tested with +20) are written in the loaded file, so the changes done are slowing logexpert down or are not very performant.

Hirogen:
after a little bit more testing, it seems that only this branch is affected by the slow down, so in this branch I'm unable to work on a file that has 20 loglines written into it every 10ms, the UI just freezes

RandallFlagg:
> I'll tested it a little bit
> 
> this is version 1.12 version and your branch both have the same result 10 ms writes are impossible to handle currently, logexpert just freezes. 20 lines per 50 ms are impossible to handle currently, logexpert just freezes. 10 lines per 50 ms can be handled but i slows down considerably, but still you can use the ui. <10 lines per 50ms easily handled.
> 
> I tested it also with the version 1.9.0 and 10 ms writes are easily handled, no slowdown, regardless of how many lines (tested with +20) are written in the loaded file, so the changes done are slowing logexpert down or are not very performant.

Now I tested this with the network per your suggestion and the UI doesn't get updated. If I open it as a regular file (which is how I originally tested it) it seems to work well. Can you tell me if the behavior in the file mode is working as intended? I will check in the meantime the network issue.

RandallFlagg:
> > Now I tested this with the network per your suggestion and the UI doesn't get updated. If I open it as a regular file (which is how I originally tested it) it seems to work well. Can you tell me if the behavior in the file mode is working as intended? I will check in the meantime the network issue.
> 
> not sure what you mean :D, I just opened the file and let a second program write into that file, to simulate a concurrently written logfile, and the current implementation starts to slow down when writing 20 lines per 10 ms into the opened file.

This is what I did. Tell me if you did something else:
1. I run your test app you provided
2. I went into log expert and opened it as a file
3. I saw lines updating
4. I closed it and opened it as a URL
5. I saw lines updating

I did this test a few times but the edges were: 1ms and 5000ms.

Hirogen:
thanks, I will test with the updated changes

RandallFlagg:
@Hirogen I get an exception when trying to open an SFTP. I get that Renci.SshNet(or something like that) is missing when trying to access SFTP. Can you check it on your system with a clean repo and tell me what I am missing? I can't seem to find it in my source dir. I can try to locate it at other places but I don't think that it is the right way to go.

Hirogen:
> @Hirogen I get an exception when trying to open an SFTP. I get that Renci.SshNet(or something like that) is missing when trying to access SFTP. Can you check it on your system with a clean repo and tell me what I am missing? I can't seem to find it in my source dir. I can try to locate it at other places but I don't think that it is the right way to go.

will check in the evening

Hirogen:
> > > Now I tested this with the network per your suggestion and the UI doesn't get updated. If I open it as a regular file (which is how I originally tested it) it seems to work well. Can you tell me if the behavior in the file mode is working as intended? I will check in the meantime the network issue.
> > 
> > 
> > not sure what you mean :D, I just opened the file and let a second program write into that file, to simulate a concurrently written logfile, and the current implementation starts to slow down when writing 20 lines per 10 ms into the opened file.
> 
> This is what I did. Tell me if you did something else:
> 
> 1. I run the test app you provided
> 2. I went into log expert and opened it as a file
> 3. I saw lines updating
> 4. I closed the file
> 5. I opened the file as a URL
> 6. I saw lines updating
> 
> The fix I did today is for steps 5 and 6. I did this test a few times, but the edges were: 1ms and 5000ms.

I tested it again, the problem starts occuring when you open the filter dialog and try to filter the file thats currently written, then the UI starts to freeze, until then, it was not a problem, only after I opened the filter dialog.

Frozen:
image

As long as the dialog is closed, the UI is fine. So something with the filter dialog is at play.

Also I was running that on my Home PC, on an M2, so it's definitely not an I/O problem of the file system, but rather an update / UI problem. Maybe someting in the LogEventWorker function, in LogWindowPrivate.cs since I got an ObjectDisposedException there, when I closed Logexpert, so it seems the UpdateGrid delegate is slower, or something in there.

image
With about 2.5 million lines, the UI is unresponsive


Currently checking also the SFTP Problem you mentioned

Hirogen:
> @Hirogen I get an exception when trying to open an SFTP. I get that Renci.SshNet(or something like that) is missing when trying to access SFTP. Can you check it on your system with a clean repo and tell me what I am missing? I can't seem to find it in my source dir. I can try to locate it at other places but I don't think that it is the right way to go.

I dont get that error, everything is fine with me, have you cleared bin / obj folders?

Hirogen:
I made a lot of improvements in the logfilereader, please check them out

Issue #403, opened 2025-06-01
Null Checks need to be added to Util.cs


Hirogen:
Null Checks need to be added to Util.cs

PR #412, opened 2025-06-06
Added Null Checks to Util.cs (issue 403)

B0risev1ch:
Added Null Checks (https://github.com/LogExperts/LogExpert/issues/403)

### DIFF
--- a/src/LogExpert.Core/Classes/Util.cs
+++ b/src/LogExpert.Core/Classes/Util.cs
@@ -11,47 +11,14 @@ public class Util
 {
     #region Public methods
 
-    public static string GetNameFromPath (string fileName)
-    {
-        var i = fileName.LastIndexOf('\\');
+    public static string GetNameFromPath (string fileName) => fileName is null ? string.Empty : Path.GetFileName(fileName);
 
-        if (i < 0)
-        {
-            i = fileName.LastIndexOf('/');
-        }
 
-        if (i < 0)
-        {
-            i = -1;
-        }
+    public static string StripExtension (string fileName) => fileName is null ? string.Empty : Path.GetFileNameWithoutExtension(fileName);
 
-        return fileName[(i + 1)..];
-    }
 
-   
-    public static string StripExtension (string fileName)
-    {
-        ArgumentNullException.ThrowIfNull(fileName);
-        var i = fileName.LastIndexOf('.');
+    public static string GetExtension (string fileName) => fileName is null ? string.Empty : Path.GetExtension(fileName).TrimStart('.');
 
-        if (i < 0)
-        {
-            i = fileName.Length - 1;
-        }
-
-        return fileName[..i];
-    }
-
-    
-    public static string GetExtension (string fileName)
-    {
-        ArgumentNullException.ThrowIfNull(fileName);
-        var i = fileName.LastIndexOf('.');
-
-        return i < 0 || i >= fileName.Length - 1
-            ? string.Empty
-            : fileName[(i + 1)..];
-    }
 
 
     public static string GetFileSizeAsText (long size)
@@ -106,9 +73,15 @@ public class Util
 
     public static int DamerauLevenshteinDistance (string src, string dest)
     {
-        ArgumentNullException.ThrowIfNull(src);
-        ArgumentNullException.ThrowIfNull(dest);
-        
+        if (dest is null || dest.Length == 0)
+        {
+            return 0;
+        }
+        if (src is null || src.Length == 0)
+        {
+            return int.MaxValue;
+        }
+
         var d = new int[src.Length + 1, dest.Length + 1];
         int i, j, cost;
         var str1 = src.ToCharArray();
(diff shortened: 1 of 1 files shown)


Hirogen:
Thanks for your contribution, but be aware and double check again please, if you throw the exception, the calling function might not handle the exception, this will lead to unexpected crashes.

Please check if all the thrown exception are handled correctly in the calling functions and if not try to implement it, but not just with a "try-catch"

RandallFlagg:
@Hirogen all the exception throwing in the commit are valid as they do exactly the same as before. The only difference now is that it is explicitly in the code and in case of a debug it is easier to see the culprit.

Hirogen:
> @Hirogen all the exception throwing in the commit are valid as they do exactly the same as before. The only difference now is that it is explicitly in the code and in case of a debug it is easier to see the culprit.

that is true, the problem I have is, it changes nothing really, its the status quo, NULL shouldn't even be possible in this situations and if it is, then something other is wrong, and this should be avoided in the first place.

The only thing this change does, is remove the warnings, but it will not fix the underlying problem, that there could be null references at a place where null references should not be.

So it will only move the goal post to another position, but not fix anything.

NULL, is always the worst case scenario and should be avoided at all times.

B0risev1ch:
Thank you to the project maintainers and contributors for your feedback — and apologies for the delay in addressing this commit.

This commit proposes the following updates:

1. New utility methods to handle file name processing in a more robust and concise way:
  ```
   public static string GetNameFromPath(string fileName) =>
       fileName is null ? string.Empty : Path.GetFileName(fileName);

   public static string StripExtension(string fileName) =>
       fileName is null ? string.Empty : Path.GetFileNameWithoutExtension(fileName); // Fixes incorrect behavior for files with no extensions (e.g., https://dotnetfiddle.net/hpUyZd)

   public static string GetExtension(string fileName) =>
       fileName is null ? string.Empty : Path.GetExtension(fileName).TrimStart('.');
```
3. Substitution cost in YetiLevenshtein(string s1, string s2) now treats null strings as empty strings;
4. Improved null handling in DamerauLevenshteinDistance(string src, string dest):
Now gracefully handles null or empty search terms.

Returns int.MaxValue when the source is null/empty (i.e., there's nothing to search _in_),

Returns 0 when the search term is null/empty (i.e., logically matches everything).

5. Update to GetWordFromPos(int xPos, string text, Graphics g, Font font):

Null checks for text, g, and font are marked with a // TODO: since this method is currently unused in the project.

These changes aim to improve robustness and readability while aligning with best practices around null safety (I hope). Let me know if any further adjustments are needed.

File: src/LogExpert.Core/Classes/Util.cs [hunk 1]
```diff
@@ -11,47 +11,14 @@ public class Util
...
 
-    public static string GetNameFromPath (string fileName)
-    {
-        var i = fileName.LastIndexOf('\\');
-
-        if (i < 0)
-        {
-            i = fileName.LastIndexOf('/');
-        }
+    public static string GetNameFromPath (string fileName) => fileName is null ? string.Empty : Path.GetFileName(fileName);
```
RandallFlagg, line 14: Why not throw an exception in case fileName is null? Why hide it?

src/LogExpert.Core/Classes/Util.cs
```diff
@@ -11,47 +11,14 @@ public class Util
...
 
-        if (i < 0)
-        {
-            i = -1;
-        }
 
-        return fileName[(i + 1)..];
-    }
+    public static string StripExtension (string fileName) => fileName is null ? string.Empty : Path.GetFileNameWithoutExtension(fileName);
```
RandallFlagg, line 17: Why not throw an exception in case fileName is null? Why hide it?

src/LogExpert.Core/Classes/Util.cs
```diff
@@ -11,47 +11,14 @@ public class Util
...
-    public static string StripExtension (string fileName)
-    {
-        ArgumentNullException.ThrowIfNull(fileName);
-        var i = fileName.LastIndexOf('.');
 
-        if (i < 0)
-        {
-            i = fileName.Length - 1;
-        }
+    public static string GetExtension (string fileName) => fileName is null ? string.Empty : Path.GetExtension(fileName).TrimStart('.');
```
RandallFlagg, line 20: Why not throw an exception in case fileName is null? Why hide it?

src/LogExpert.Core/Classes/Util.cs
```diff
@@ -106,9 +73,15 @@ public static bool TestFilterCondition (FilterParams filterParams, ILogLine line
 
     public static int DamerauLevenshteinDistance (string src, string dest)
     {
-        ArgumentNullException.ThrowIfNull(src);
-        ArgumentNullException.ThrowIfNull(dest);
-        
+        if (dest is null || dest.Length == 0)
```
RandallFlagg, line 76: Why to hide? so we will know how to prevent this in the future

Hirogen:
I added most of your changes in another PR, while making performance optimizations, please review them at your leisure and if you have a few other changes you like to make, open a new PR, thanks for you contribution.

PR #429, opened 2025-06-18
Log file reader

RandallFlagg:
@Hirogen can you please look at the change I did in `LogWindowPublic`?
I am a bit confused as why everything is working by me. Am I missing a scenario where that change is not good? This created a huge resource and performance impact on my system.


Hirogen:
> @Hirogen can you please look at the change I did in `LogWindowPublic`? I am a bit confused as why everything is working by me. Am I missing a scenario where that change is not good? This created a huge resource and performance impact on my system.

Seems to work normal on my system, 6 files with 500mb open and uses about 5gig on memory and a little bit of processor time

image


But it's not possible to close a file while the filter is running, it will throw a cross thread exception, this is probably because the StatuslineEventWorker has been started by a Task and is now called again by another thread.

I also get 
image

but I hope this is due to the missing catch()

src/LogExpert.Core/Classes/Log/LogfileReader.cs
```diff
@@ -755,43 +779,74 @@ private ILogFileInfo AddFile (string fileName)
...
+                _logger.Error("Cannot find buffer for line {0}, file: {1}{2}", lineNum, _fileName, IsMultiFile ? " (MultiFile)" : "");
+                return null;
             }
 
-            _disposeLock.DowngradeFromWriterLock(ref cookie);
-        }
+            _disposeLock.AcquireReaderLock(Timeout.Infinite);
+            try
+            {
+                ct.ThrowIfCancellationRequested();
```
Hirogen, line 818: I know it's just a draft, but the try, finally is imo worse then the old design

Also I'm confused why so many ct.ThrowIfCancellationRequested();

Hirogen:
I made several improvements to the performance, please check them out, currently I'm working on new performance enhancements.

PR #447, opened 2025-07-10
446 highlighting and action trigger dialog too big for laptop screens

### LINKED ISSUE #446: "Highlighting and action trigger" dialog too big for Laptop screens
# Context
When logexpert is used on laptop screens (eg. 1920x 1080), the **Highlight and action trigger** dialog is higher than the max. screen height when opening. Minimizing the dialog helps a little bit, but it does not get smaller than 1080px. 

# Proposal
- Embed all controls of the dialog in a scroll panel that shows the scrollbars only if necessary. 
- Reduce the initial height to at least 1080px

Hirogen:
Change UI Size to 660x800 minimum so it fits on a 1080p screen with 125% DPI
Beware everything above (150% or 175%) will have the same effect as before.


Hirogen:
Snapshot for Testing purposes
[Logexpert.Debug.Snapshot.zip](https://github.com/user-attachments/files/21156516/Logexpert.Debug.Snapshot.zip)

stprograms:
Looks feasible on a 1080px monitor with scaling of 125%.

PR #455, opened 2025-07-21
448 allow only 1 instance setting does nothing

### LINKED ISSUE #448: "Allow only 1 Instance" setting does nothing
Even when checked:

Image

Opening more files results in multiple instances of LogExpert instead of multiple tabs.

Seems like the setting has no impact at all.

Tested on V1.20.0

Hirogen:
Rework of Allow only one Instance

Lock Instance and "Allow only one Instance" have similar functionality, will be moved together


src/LogExpert.UI/Dialogs/SettingsDialog.cs
```diff
@@ -210,10 +216,12 @@ private void SaveMultifileData ()
 
     private void OnBtnToolClickInternal (TextBox textBox)
     {
-        OpenFileDialog dlg = new();
-        dlg.InitialDirectory = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles);
+        OpenFileDialog dlg = new()
+        {
+            InitialDirectory = Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles)
```
RandallFlagg, line 288: I know you didn't do this change but why is the initial folder Program Files?

src/LogExpert.UI/Dialogs/SettingsDialog.cs
```diff
@@ -210,10 +216,12 @@ private void SaveMultifileData ()
...
+        };
 
-        if (string.IsNullOrEmpty(textBox.Text) == false)
+        if (!string.IsNullOrEmpty(textBox.Text))
```
RandallFlagg, line 291: I think IsNullOrWhiteSpace is better here

src/LogExpert.UI/Dialogs/SettingsDialog.cs
```diff
@@ -652,22 +648,13 @@ private void OnBtnOkClick (object sender, EventArgs e)
...
-        }
-        else if (radioButtonSessionApplicationStartupDir.Checked)
-        {
-            Preferences.SaveLocation = SessionSaveLocation.ApplicationStartupDir;
-        }
-        else
-        {
-            Preferences.SaveLocation = SessionSaveLocation.SameDir;
-        }
+        Preferences.SaveLocation = radioButtonsessionSaveDocuments.Checked
```
RandallFlagg, line 749: Here is another option. See if you like it:
```cs
Preferences.SaveLocation = radioButtonsessionSaveDocuments.Checked switch
{
    true  => SessionSaveLocation.DocumentsDir,
    _ when radioButtonSessionSaveOwn.Checked               => SessionSaveLocation.OwnDir,
    _ when radioButtonSessionApplicationStartupDir.Checked => SessionSaveLocation.ApplicationStartupDir,
    _                                                      => SessionSaveLocation.SameDir
};

```

PR #469, opened 2025-09-24
Fix ArgumentException when FontName in setting file is not valid

### LINKED ISSUE #468: Can't start LogExpert if font in settings file is not found
After restoring settings.json file from a backup on a new installation, I get following error in a MessageBox upon start:
 > Cannot open connection to first instance (Font 'Hack NF' cannot be found.)
Closing the message exits program.

I get that the "Cannot open connection to first instance" comes from implementation of the "Allow only one instance" option but missing font or in this case changed font name should not cause an exception or block startup.

This is LogExpert version 1.20 but same issue is present after building from the Development branch.

The line from the NLog log file and the actual exception hidden behind the MessageBox:
```log
Mutex error, giving up: Font 'Hack NF' cannot be found. System.ArgumentException System.ArgumentException: Font 'TestMissingFont' cannot be found.
   at System.Drawing.FontFamily.CreateFontFamily(String name, FontCollection fontCollection)
   at System.Drawing.FontFamily..ctor(String name)
   at LogExpert.Dialogs.BookmarkWindow.SetFont(String fontName, Single fontSize) in C:\Development\LogExpert\src\LogExpert.UI\Dialogs\BookmarkWindow.cs:line 229
   at LogExpert.Dialogs.BookmarkWindow.PreferencesChanged(String fontName, Single fontSize, Boolean setLastColumnWidth, Int32 lastColumnWidth, SettingsFlags flags) in C:\Development\LogExpert\src\LogExpert.UI\Dialogs\BookmarkWindow.cs:line 157
   at LogExpert.UI.Controls.LogTabWindow.LogTabWindow.InitBookmarkWindow() in C:\Development\LogExpert\src\LogExpert.UI\Dialogs\LogTabWindow\LogTabWindow.cs:line 684
   at LogExpert.UI.Controls.LogTabWindow.LogTabWindow.InitToolWindows() in C:\Development\LogExpert\src\LogExpert.UI\Dialogs\LogTabWindow\LogTabWindow.cs:line 661
   at LogExpert.UI.Controls.LogTabWindow.LogTabWindow..ctor(String[] fileNames, Int32 instanceNumber, Boolean showInstanceNumbers, IConfigManager configManager) in C:\Development\LogExpert\src\LogExpert.UI\Dialogs\LogTabWindow\LogTabWindow.cs:line 177
   at LogExpert.UI.Extensions.LogWindow.AbstractLogTabWindow.Create(String[] fileNames, Int32 instanceNumber, Boolean showInstanceNumbers, IConfigManager configManager) in C:\Development\LogExpert\src\LogExpert.UI\Extensions\LogWindow\AbstractLogTabWindow.cs:line 15
   at LogExpert.Program.Main(String[] args) in C:\Development\LogExpert\src\LogExpert\Program.cs:line 91
```

That could be solved by catching a more specific exception in the try..catch statement. Also applies to other places in the codebase where this is prevalent.

Another similar issue with settings JSON deserialization where settings are silently reset to defaults. That leads to loss of settings file contents and subsequent user frustration. As far as I can tell, columnizer references can't be instantiated properly. In some places the `[JsonIgnore]` property attribute (as in [FilterParams.CurrentColumnizer in FilterParams.cs](https://github.com/LogExperts/LogExpert/blob/0c1ed7f9fc7a4720113698df99f1f2d29fe9e996/src/LogExpert.Core/Classes/Filter/FilterParams.cs#L55)) is actually from `System.Text.Json.Serialization` but settings serialization is done with `Newtonsoft.Json` so the attribute is ignored.

I think I've solved the latter issue and could make a PR. Just need to do some manual tests but it's already 1am here.

AndisGrossteins:
Fixes LogExperts/LogExpert#468

The check for invalid font name and falling back to default font might be better placed in LogExpert.Config.ConfigManager but I've added it in LogWindow where the exception was thrown blocking application startup.

### DIFF
--- a/src/LogExpert/Config/ConfigManager.cs
+++ b/src/LogExpert/Config/ConfigManager.cs
@@ -206,13 +206,14 @@ public class ConfigManager : IConfigManager
 
             try
             {
-                var fontFamily = new FontFamily(settings.Preferences.FontName);
+                using var fontFamily = new FontFamily(settings.Preferences.FontName);
                 settings.Preferences.FontName = fontFamily.Name;
             }
             catch (ArgumentException ex)
             {
-                _logger.Warn($"Specified font '{settings.Preferences.FontName}' not found. Falling back to default: '{FontFamily.GenericMonospace.Name}'. {ex.Message}");
-                settings.Preferences.FontName = FontFamily.GenericMonospace.Name;
+                var genericMonospaceFont = FontFamily.GenericMonospace.Name;
+                _logger.Warn($"Specified font '{settings.Preferences.FontName}' not found. Falling back to default: '{genericMonospaceFont}'.");
+                settings.Preferences.FontName = genericMonospaceFont;
             }
 
             if (settings.Preferences.ShowTailColor == Color.Empty)
--- a/src/LogExpert/Config/ConfigManager.cs
+++ b/src/LogExpert/Config/ConfigManager.cs
@@ -209,7 +209,7 @@ public class ConfigManager : IConfigManager
                 using var fontFamily = new FontFamily(settings.Preferences.FontName);
                 settings.Preferences.FontName = fontFamily.Name;
             }
-            catch (ArgumentException ex)
+            catch (ArgumentException)
             {
                 var genericMonospaceFont = FontFamily.GenericMonospace.Name;
                 _logger.Warn($"Specified font '{settings.Preferences.FontName}' not found. Falling back to default: '{genericMonospaceFont}'.");
--- a/src/LogExpert/Config/ConfigManager.cs
+++ b/src/LogExpert/Config/ConfigManager.cs
@@ -204,6 +204,17 @@ public class ConfigManager : IConfigManager
 
             settings.FileColors ??= [];
 
+            try
+            {
+                var fontFamily = new FontFamily(settings.Preferences.FontName);
+                settings.Preferences.FontName = fontFamily.Name;
+            }
+            catch (ArgumentException ex)
+            {
+                _logger.Warn($"Specified font '{settings.Preferences.FontName}' not found. Falling back to default: '{FontFamily.GenericMonospace.Name}'. {ex.Message}");
+                settings.Preferences.FontName = FontFamily.GenericMonospace.Name;
+            }
+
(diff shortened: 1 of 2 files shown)


Hirogen:
why not implement it in the configmanager? if it would be the right place?

AndisGrossteins:
> why not implement it in the configmanager? if it would be the right place?

I wasn't sure how to approach it better. Now moved it into ConfigManager.

src/LogExpert/Config/ConfigManager.cs
```diff
@@ -204,6 +204,17 @@ private Settings LoadOrCreateNew (FileInfo fileInfo)
...
             settings.FileColors ??= [];
 
+            try
+            {
+                var fontFamily = new FontFamily(settings.Preferences.FontName);
+                settings.Preferences.FontName = fontFamily.Name;
+            }
+            catch (ArgumentException ex)
+            {
+                _logger.Warn($"Specified font '{settings.Preferences.FontName}' not found. Falling back to default: '{FontFamily.GenericMonospace.Name}'. {ex.Message}");
```
Hirogen: No need to log the exception in the warning

AndisGrossteins: I agree. I'll add `using` to dispose of the FontFamily instance as suggested by Copilot.

src/LogExpert/Config/ConfigManager.cs
```diff
@@ -204,6 +204,18 @@ private Settings LoadOrCreateNew (FileInfo fileInfo)
 
             settings.FileColors ??= [];
 
+            try
+            {
+                using var fontFamily = new FontFamily(settings.Preferences.FontName);
+                settings.Preferences.FontName = fontFamily.Name;
+            }
+            catch (ArgumentException ex)
```
Hirogen: i'm a little nit picky here, but you can remove the  "ex" variable, because we don't use it anyway :)

Hirogen (dismissed):

src/LogExpert/Config/ConfigManager.cs
AndisGrossteins: Ah, right. Let me just remove it, then.

Hirogen (approved):

PR #662, opened 2026-07-23
Persistence fix on app exit/start + Highlight dialog Import/Export error fix

Pr0metheus2:
Bug Fix :
- .lxp persistence on exit when automatic saving and filter-tab restoration are enabled.
- Save the live filter UI state before shutdown teardown, so FilterParamsList is no longer written empty or stale.
- Prevent close-time tab disposal from overwriting the already saved snapshot.
Added:
- a regression test covering the snapshot behavior.

### DIFF
--- a/src/LogExpert.Resources/Resources.resx
+++ b/src/LogExpert.Resources/Resources.resx
@@ -239,7 +239,7 @@
         <value>Export Settings to file</value>
     </data>
   <data name="HighlightDialog_UI_Export_Filter" xml:space="preserve">
-        <value>Settings (*.json)|*.json|All files (*.*)</value>
+        <value>Settings (*.json)|*.json|All files (*.*)|*.*</value>
     </data>
   <data name="HighlightDialog_UI_Snippet_CopyOf" xml:space="preserve">
         <value>Copy of</value>
--- a/src/LogExpert.Resources/Resources.zh-CN.resx
+++ b/src/LogExpert.Resources/Resources.zh-CN.resx
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
   <!-- 
     Microsoft ResX Schema 
@@ -148,7 +148,7 @@
     <value>导出设置到文件</value>
   </data>
   <data name="HighlightDialog_UI_Export_Filter" xml:space="preserve">
-    <value>设置文件 (*.json)|*.json|所有文件 (*.*)</value>
+      <value>设置文件 (*.json)|*.json|所有文件 (*.*)|*.*</value>
   </data>
   <data name="HighlightDialog_UI_Snippet_CopyOf" xml:space="preserve">
     <value>复制为</value>
--- a/src/LogExpert.Tests/Controls/LogWindowPersistenceTests.cs
+++ b/src/LogExpert.Tests/Controls/LogWindowPersistenceTests.cs
@@ -1,0 +1,166 @@
+using System.Reflection;
+using System.Runtime.Versioning;
+
+using LogExpert.Core.Classes.Filter;
+using LogExpert.Core.Config;
+using LogExpert.Core.Interfaces;
+using LogExpert.UI.Controls.LogWindow;
+using LogExpert.UI.Interface;
+
+using Moq;
+
+using NUnit.Framework;
+
+namespace LogExpert.Tests.Controls;
+
+[TestFixture]
+[Apartment(ApartmentState.STA)]
+[SupportedOSPlatform("windows")]
+public sealed class LogWindowPersistenceTests : IDisposable
+{
+    private Mock<ILogWindowCoordinator> _coordinatorMock = null!;
+    private Mock<IConfigManager> _configManagerMock = null!;
+    private Settings _settings = null!;
+    private WindowsFormsSynchronizationContext? _syncContext;
+    private LogWindow _logWindow = null!;
+    private bool _disposed;
+
+    [OneTimeSetUp]
+    public void OneTimeSetUp ()
+    {
+        var dir = Path.GetDirectoryName(typeof(LogWindowPersistenceTests).Assembly.Location)!;
+        _ = PluginRegistry.PluginRegistry.Create(dir, 500);
+    }
+
+    [SetUp]
+    public void SetUp ()
+    {
+        if (SynchronizationContext.Current == null)
+        {
+            _syncContext = new WindowsFormsSynchronizationContext();
+            SynchronizationContext.SetSynchronizationContext(_syncContext);
+        }
+
+        _coordinatorMock = new Mock<ILogWindowCoordinator>();
+        _ = _coordinatorMock.Setup(c => c.ResolveHighlightGroup(It.IsAny<string?>(), It.IsAny<string?>()))
+            .Returns(new LogExpert.Core.Entities.HighlightGroup());
+        _ = _coordinatorMock.Setup(c => c.SearchParams).Returns(new LogExpert.Core.Entities.SearchParams());
+
+        _configManagerMock = new Mock<IConfigManager>();
+        _settings = new Settings();
+        _ = _configManagerMock.Setup(cm => cm.Settings).Returns(_settings);
+
+        _settings.FilterList.Add(new FilterParams
+        {
+            SearchText = "SHARED_FILTER_LIST_ENTRY",
+            IsCaseSensitive = false,
+            IsRegex = false,
+            IsFilterTail = true,
+            FuzzyValue = 0,
+            SpreadBefore = 0,
+            SpreadBehind = 0
+        });
+
+        _logWindow = new LogWindow(
+            _coordinatorMock.Object,
+            "test.log",
+            isTempFile: false,
+            forcePersistenceLoading: false,
+            configManager: _configManagerMock.Object);
+    }
+
+    [TearDown]
+    public void TearDown ()
+    {
+        _logWindow?.Dispose();
+        _syncContext?.Dispose();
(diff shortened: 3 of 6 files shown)


Pr0metheus2:
Seems error on GitHub side? Or how fix it?

image

src/LogExpert.UI/Controls/LogWindow/LogWindow.cs
```diff
@@ -5983,6 +6000,24 @@ public SessionSnapshot GatherSessionSnapshot ()
         return snapshot;
     }
 
+    private FilterParams CaptureCurrentFilterParams ()
```
Hirogen, line 6003: Duplicated Code — CaptureCurrentFilterParams() repeats the controls→FilterParams copy shape already present at ~4420–4440 (FilterSearch) and again in the dirty-check at ~4770–4795. Three sites now share the shape; one helper is warranted — though "Surgical Changes" partially excuses not touching the pre-existing sites.

src/LogExpert.Tests/Controls/LogWindowPersistenceTests.cs
```diff
@@ -0,0 +1,166 @@
...
+        _syncContext = null;
+    }
+
+    public void Dispose ()
+    {
+        Dispose(true);
+        GC.SuppressFinalize(this);
+    }
+
+    private void Dispose (bool disposing)
```
Hirogen, line 86: Duplicated Code / Speculative Generality (tests) — the test fixture implements a full Dispose(bool)/_disposed pattern that duplicates its own TearDown; possibly analyzer-driven (NUnit1032), but the guard-flag pattern is overkill either way.

src/LogExpert.Tests/Controls/LogWindowPersistenceTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+        _disposed = true;
+    }
+
+    [Test]
+    public void GatherSessionSnapshot_SavesLiveFilterState_InsteadOfSharedFilterList ()
+    {
+        SetText("filterComboBox", "LIVE_FILTER_TEXT");
+        SetText("filterRangeComboBox", "LIVE_RANGE_TEXT");
+        SetChecked("filterCaseSensitiveCheckBox", true);
```
Hirogen, line 107: Speculative Generality (tests) — SetChecked/SetValue use reflection (GetProperty("Checked")) on controls that could be typed directly.

src/LogExpert.UI/Controls/LogWindow/LogWindow.cs
```diff
@@ -168,6 +168,15 @@ internal partial class LogWindow : DockContent, ILogPaintContextUI, ILogView, IL
...
     private bool _waitingForClose;
 
+    /// <summary>
+    /// Set by the application shutdown path after a pre-save pass has already captured the
+    /// current persistence state. Prevents the per-window close handler from writing an older
+    /// snapshot after child filter tabs have already been torn down.
+    /// </summary>
+    [Browsable(false)]
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+    internal bool SkipPersistenceSaveOnClose { get; set; }
```
Hirogen, line 178: Control-flag coupling — LogTabWindow mutates SkipPersistenceSaveOnClose on children before closing them; a Close(savePersistence: false) path would avoid the mutable flag.

src/LogExpert.Tests/Controls/LogWindowPersistenceTests.cs
```diff
@@ -0,0 +1,166 @@
+using System.Reflection;
```
Hirogen, line 1: "Added: a regression test covering the snapshot behavior." — Partial. The test covers only the live-UI capture in GatherSessionSnapshot. Neither of the other two claims is tested: nothing exercises "Save the live filter UI state before shutdown teardown" (the ordering in OnLogTabWindowFormClosing), and SkipPersistenceSaveOnClose is never touched by any test.
".lxp persistence on exit/start" — the "start" half has no changes; the restore path (RestoreFilters, [LogWindow.cs:2677] already used FilterParamsList[0], so exit-side changes alone plausibly complete the fix.

src/LogExpert.Resources/Resources.resx
```diff
@@ -239,7 +239,7 @@
         <value>Export Settings to file</value>
     </data>
   <data name="HighlightDialog_UI_Export_Filter" xml:space="preserve">
-        <value>Settings (*.json)|*.json|All files (*.*)</value>
+        <value>Settings (*.json)|*.json|All files (*.*)|*.*</value>
```
Hirogen, line 242: This is a fix for export, not import! Import already has it.

src/LogExpert.Resources/Resources.zh-CN.resx
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
Hirogen, line 1: zh-CN BOM + re-indent, EOF-newline fixes in four files. Pls remove boms if they were not present before

src/LogExpert.UI/Controls/LogWindow/LogWindow.cs
```diff
@@ -905,7 +914,10 @@ private void OnLogWindowClosing (object sender, CancelEventArgs e)
             return;
         }
 
-        SavePersistenceData(false);
+        if (!SkipPersistenceSaveOnClose)
```
Hirogen, line 917: "Prevent close-time tab disposal from overwriting the already saved snapshot." — Works for the guarded path (pre-save runs before RemoveAndDisposeLogWindow, and OnLogWindowClosing at [LogWindow.cs:917] (skips the re-save). But the flag is set-only and never reset — if FormClosing is ever cancelled after the handler runs, surviving windows silently skip persistence on all later manual tab closes. Also Reload() ([LogWindow.cs:7420] and ReloadNewFile() call SavePersistenceData(false) unguarded — a rollover firing mid-shutdown could still overwrite the snapshot. Narrow races, worth noting.

src/LogExpert.UI/Controls/LogWindow/LogWindow.cs
```diff
@@ -168,6 +168,15 @@ internal partial class LogWindow : DockContent, ILogPaintContextUI, ILogView, IL
 
     private bool _waitingForClose;
 
+    /// <summary>
+    /// Set by the application shutdown path after a pre-save pass has already captured the
+    /// current persistence state. Prevents the per-window close handler from writing an older
+    /// snapshot after child filter tabs have already been torn down.
+    /// </summary>
+    [Browsable(false)]
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
```
Hirogen, line 177: [Browsable(false)]/[DesignerSerializationVisibility] on SkipPersistenceSaveOnClose are speculative decoration; LogWindow is never designer-hosted with this property.

Hirogen (changes requested):

Hirogen:
> Seems error on GitHub side? Or how fix it?
> image

probably because the github actions cant write on the branch, development branches are normally protected, thats why a separate issue branch should be created and then a PR into the dev branch, to avoid it.

src/LogExpert.UI/Controls/LogWindow/LogWindow.cs
Pr0metheus2: Thanks. I agree the duplication is real, but I kept this commit "surgical" and only touched the new capture path.

src/LogExpert.UI/Controls/LogWindow/LogWindow.cs
Pr0metheus2: Agree that a Close(savePersistence: false) would be cleaner. I used the existing flag here to avoid widening the close-path change in this commit and to keep the behavior isolated.

src/LogExpert.UI/Controls/LogWindow/LogWindow.cs
Pr0metheus2: Could be a concern. The flag is only set in the app shutdown cascade, immediately before the pre-save and disposal sequence. I don’t see a normal path where it remains active after that close flow completes.
This is relevant if the close cascade aborts or throws before disposal completes. LogTabWindow does swallow these exceptions in its close handler.

src/LogExpert.UI/Controls/LogWindow/LogWindow.cs
Pr0metheus2: It looks unnecessary because this is not a designer-driven surface, but it is harmless metadata, not a correctness issue. I would treat it as cleanup, not a defect. They are not behavior-changing.

src/LogExpert.Tests/Controls/LogWindowPersistenceTests.cs
Pr0metheus2: Mostly style noise, not a correctness issue.

src/LogExpert.Tests/Controls/LogWindowPersistenceTests.cs
Pr0metheus2: Yes, mostly style noise, but it is not a product bug.

src/LogExpert.Tests/Controls/LogWindowPersistenceTests.cs
Pr0metheus2: You are right, “exit/start” is only partially backed by tests. The fix is primarily on the save/exit side, not on restore logic.

Pr0metheus2:
If you recommend, I can close this pull request and start over again with small individual steps.

PR #663, opened 2026-07-24
Fix: "Highlight and action triggers" Export function

Pr0metheus2:
"Highlighting and action triggers" Export function was causing unhandled error exception. The filter now uses valid description|pattern pairs, so exporting highlight groups no longer throws an error dialog.

image

### DIFF
--- a/src/LogExpert.Resources/Resources.de.resx
+++ b/src/LogExpert.Resources/Resources.de.resx
@@ -233,7 +233,7 @@
 		<value>Exportieren der Einstellungen in eine Datei</value>
 	</data>
   <data name="HighlightDialog_UI_Export_Filter" xml:space="preserve">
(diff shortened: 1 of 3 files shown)


Hirogen (approved):

Hirogen:
hmm you still try to merge from development (protected) to development (protected) thats the reason, the hashes can not be updated in your branch, because its protected

Pr0metheus2:
I have no idea what should I setup in my repository to get it passed...

Hirogen:
The CI failure isn't your change — the workflow itself was broken for fork PRs (checkout resolved your branch name against this repo instead of your fork, and the hash auto-commit step can't push from a fork's read-only token). Fixed in #667. Once that's merged, please rebase or push any new commit to re-trigger the build — a plain re-run won't pick up the fixed workflow.

Pr0metheus2:
OK

PR #685, opened 2026-07-29
Fix: Open tabs order on next start

Pr0metheus2:
LogExpert does not remember open tabs order from last instance. When app starts, tabs are restored in different order from last state. This change should fix mentioned issue.

### Before proposed change:
Initial tab order:
image
Tab order after close/open:
image

### After proposed change:
Initial tab order:
image
Tab order after close/open:
image

### DIFF
--- a/src/LogExpert.UI/Services/TabControllerService/TabController.cs
+++ b/src/LogExpert.UI/Services/TabControllerService/TabController.cs
@@ -458,15 +458,16 @@ internal class TabController : ITabController
     /// Gets all LogWindow instances from the DockPanel's Contents collection.
     /// </summary>
     /// <returns>Read-only list of all LogWindows in the DockPanel</returns>
-    public IReadOnlyList<LogWindow> GetAllWindowsFromDockPanel ()
-    {
-        return !_initialized || _dockPanel == null
-            ? []
-            : (IReadOnlyList<LogWindow>)_dockPanel.Contents
-            .OfType<LogWindow>()
-            .ToList()
-            .AsReadOnly();
-    }
+	public IReadOnlyList<LogWindow> GetAllWindowsFromDockPanel ()
+	{
+		return !_initialized || _dockPanel == null
+			? []
+			: _dockPanel.Panes
+				.Where(pane => pane.DockState == DockState.Document)
+				.SelectMany(pane => pane.Contents.OfType<LogWindow>())
+				.ToList()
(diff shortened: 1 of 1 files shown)


src/LogExpert.UI/Services/TabControllerService/TabController.cs
```diff
@@ -458,15 +458,16 @@ protected virtual void Dispose (bool disposing)
...
-            .ToList()
-            .AsReadOnly();
-    }
+	public IReadOnlyList<LogWindow> GetAllWindowsFromDockPanel ()
+	{
+		return !_initialized || _dockPanel == null
+			? []
+			: _dockPanel.Panes
+				.Where(pane => pane.DockState == DockState.Document)
+				.SelectMany(pane => pane.Contents.OfType<LogWindow>())
```
Hirogen, line 467: Why first filter only on DockState.Document, there could be free floating tabs.
Secondly, why SelectMany, if in the original, only the Contents.OfType<LogWindow> was preselected... now you added 2 LinqQueries, depending on the sice of the list this and the content this could be a performance issue

Pr0metheus2: I agree with your concern. Can you check this code, please? Tabs restored fine.

```
    public IReadOnlyList<LogWindow> GetAllWindowsFromDockPanel ()
    {
        if (!_initialized || _dockPanel == null)
        {
            return [];
        }

        List<LogWindow> windows = [];

        // DisplayingContents is the order used by the visible tab strip.
        foreach (DockPane pane in _dockPanel.Panes)
        {
            for (var index = 0; index < pane.DisplayingContents.Count; index++)
            {
                if (pane.DisplayingContents[index] is LogWindow logWindow)
                {
                    windows.Add(logWindow);
                }
            }
        }

        return windows.AsReadOnly();
    }
```

Hirogen: hmm on further thinking the linq-query is probably fine, since nobody will have hundreds open tabs with logfiles, you just need to keep in mind that pane.DockState == DockState.Document is not the only state a Tab can have! so floating states would be swallowed. In [LogTabWindow.cs:836] AddLogWindow sets logWindow.DockAreas = DockAreas.Document | DockAreas.Float, so those 2 need be kept in mind

Hirogen (approved):

PR #691, opened 2026-07-30
fix: replace the last-open-files list when a window closes (#686)

### LINKED ISSUE #686: LastOpenFilesList is never cleared when startup files are passed or "Open last files" is disabled
### Summary

`LastOpenFilesList` is only cleared inside one branch of `LoadStartupFiles`. Two common startup paths skip that clear, after which every application close *appends* the current tabs to the stale list instead of replacing it. The list grows without bound, and once it is read again the user gets tabs they closed sessions ago, opened before the current ones.

### Steps to reproduce

1. Enable **Open last files** in the preferences.
2. Start LogExpert normally, open `a.log` and `b.log`, close the app.
3. Start LogExpert by double-clicking `c.log` in Explorer (i.e. with a command-line file argument). Close the app.
4. Start LogExpert normally again.

**Expected:** only `c.log` is restored.
**Actual:** `a.log`, `b.log` and `c.log` are restored, with the stale files first. `settings.json` now holds five entries for three files.

Repeating steps 3–4 keeps growing the list.

The same happens with **Open last files** disabled: the list is never read and never cleared, so it accumulates silently. Enabling the preference later opens a flood of stale tabs.

### Root cause

`src/LogExpert.UI/Services/FileOperationService/FileOperationService.cs`

`ClearLastOpenFilesList()` is called from exactly one place — line 353, inside the `if (_configManager.Settings.Preferences.OpenLastFiles)` branch of `LoadStartupFiles`, after the reopen loop:

```csharp
public void LoadStartupFiles (IList<string> lastOpenFiles, string[]? startupFileNames)
{
    if (startupFileNames != null && startupFileNames.Length > 0)
    {
        _ = LoadFilesWithOption(startupFileNames, false);
        return;                                   // <-- no clear
    }

    if (_configManager.Settings.Preferences.OpenLastFiles)   // <-- no clear when false
    {
        foreach (var name in lastOpenFiles.Where(filename => !string.IsNullOrEmpty(filename)))
        {
            _ = AddFileTab(new FileTabRequest { FileName = name });
        }

        _configManager.ClearLastOpenFilesList();
    }
}
```

`SaveLastOpenFilesList` (line 330) only ever calls `Add`, so it depends on that clear having happened:

```csharp
public void SaveLastOpenFilesList ()
{
    foreach (var logWin in _tabController.GetAllWindowsFromDockPanel().Where(logwin => !logwin.IsTempFile))
    {
        _configManager.Settings.LastOpenFilesList.Add(logWin.GivenFileName);
    }
}
```

It is invoked unconditionally from `OnLogTabWindowFormClosing` (`src/LogExpert.UI/Dialogs/LogTabWindow/LogTabWindow.cs:1759`), so save and clear are not symmetric.

Duplicate entries do not create duplicate tabs — `AddFileTab` resolves an existing window by normalized full path (line 66) and activates it instead — so the visible symptoms are unwanted tabs and an ever-growing settings list, not duplicated tabs.

### Suggested fix

Make the save total rather than additive, so it no longer depends on a clear elsewhere:

```csharp
public void SaveLastOpenFilesList ()
{
    _configManager.ClearLastOpenFilesList();

    foreach (var logWin in _tabController.GetAllWindowsFromDockPanel().Where(logwin => !logwin.IsTempFile))
    {
        _configManager.Settings.LastOpenFilesList.Add(logWin.GivenFileName);
    }
}
```

Alternative: move `ClearLastOpenFilesList()` in `LoadStartupFiles` so it runs on every path including the command-line early return. The first option is preferable — it makes the method self-contained and idempotent regardless of how startup went.

### Notes

- `src/LogExpert.Tests/Services/FileOperationServiceTests.cs` asserts `ClearLastOpenFilesList` call counts (`Times.Never` / `Times.Once`) around lines 803–866 and in the `SaveLastOpenFilesList_*` tests; those expectations need updating with either fix.
- Independent of #685, which addresses the *order* of the entries this method writes, not their staleness.

Hirogen:
Fixes #686.

SaveLastOpenFilesList only ever appended, while the clear lived in one branch of LoadStartupFiles. Starting with command-line files, or with Open last files off, skipped that clear — so each close appended the current tabs to a stale list, which grew without bound and restored tabs closed sessions ago.

The closing window now owns the list. The first window to close replaces it; windows closing after it append, so a multi-window session is still restored as a whole rather than only the last window to close. The clear in LoadStartupFiles is removed, so a crash mid-session no longer wipes the set of files to restore.

Behaviour notes:

- Within one long-lived process, closing several windows in sequence unions their files — same as before this change, and bounded by process lifetime instead of accumulating across restarts.
- The clear-then-add pair is not atomic (only the clear takes _loadSaveLock), unchanged from before and only reachable if two windows close at the same instant.

Tests: replace-on-first-close, append-on-second-window, clear when only temp files are open, and load-keeps-the-list.

### DIFF
--- a/src/LogExpert.Tests/Services/FileOperationServiceTests.cs
+++ b/src/LogExpert.Tests/Services/FileOperationServiceTests.cs
@@ -54,6 +54,10 @@ internal class FileOperationServiceTests : IDisposable
 
         _settings = new Settings();
         _ = _configManagerMock.Setup(cm => cm.Settings).Returns(_settings);
+        _ = _configManagerMock.Setup(cm => cm.ClearLastOpenFilesList()).Callback(() => _settings.LastOpenFilesList.Clear());
+
+        // Process-wide state: every test starts as if no window has saved its files yet
+        FileOperationService.LastOpenFilesListReplaced = false;
         _ = _pluginRegistryMock.Setup(pr => pr.RegisteredColumnizers).Returns([]);
 
         _factoryCalls = [];
@@ -799,9 +803,6 @@ internal class FileOperationServiceTests : IDisposable
         // Assert — startup files are loaded (via LoadFilesWithOption → AddFileTab)
         Assert.That(_factoryCalls, Has.Count.EqualTo(1));
         Assert.That(_factoryCalls[0].Request.FileName, Is.EqualTo("startup.log"));
-
-        // Last-open files should NOT be loaded
-        _configManagerMock.Verify(cm => cm.ClearLastOpenFilesList(), Times.Never);
     }
 
     [Test]
@@ -818,7 +819,6 @@ internal class FileOperationServiceTests : IDisposable
         Assert.That(_factoryCalls, Has.Count.EqualTo(2));
         Assert.That(_factoryCalls[0].Request.FileName, Is.EqualTo("file1.log"));
         Assert.That(_factoryCalls[1].Request.FileName, Is.EqualTo("file2.log"));
-        _configManagerMock.Verify(cm => cm.ClearLastOpenFilesList(), Times.Once);
     }
 
     [Test]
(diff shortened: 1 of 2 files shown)


Pr0metheus2:
When app closes, it looks like it wipes the set of files to restore. Older version (I think from 2 days ago) does remember this.

###Before close:
image

###After close:
image

Hirogen:
> Older version

interesting, should only wipe it if no files are open, probably a bug

Pr0metheus2:
Can you check this edit if properly fixed? In my test, the bug is not present anymore. If correct, I can create Pull request. **TabController.cs**:462

```
	public IReadOnlyList<LogWindow> GetAllWindowsFromDockPanel ()
	{
	  if (!_initialized || _dockPanel == null)
	  {
		  return [];
	  }

	  List<LogWindow> windows = [];

	  foreach (DockPane pane in _dockPanel.Panes)
	  {
		  var displayingContents = pane.DisplayingContents;

		  for (int index = 0; index < displayingContents.Count; index++)
		  {
			  if (displayingContents[index] is LogWindow logWindow)
			  {
				  windows.Add(logWindow);
			  }
		  }
	  }

      return windows.AsReadOnly();
	}
```


image

Hirogen:
just open a new pr i will review that one, its easier, than reviewing code snippets in comments

PR #692, opened 2026-08-03
fix/SaveLastOpenFilesList()

Pr0metheus2:
Fix for clearing "LastOpenFilesList" on app exit. 
When app closes, it looks like it wipes the set of files to restore.

Before exit:
image

After exit:
image

Also fixes two other places (both places are affected by this fix and should now work correctly):
  - **Close other tabs** (`LogTabWindow.cs:2326`)
  - **Session-save file list** (`LogTabWindow.cs:2410`)

### DIFF
--- a/src/LogExpert.Tests/Services/TabControllerTests.cs
+++ b/src/LogExpert.Tests/Services/TabControllerTests.cs
@@ -40,7 +40,8 @@ internal class TabControllerTests : IDisposable
         {
             Dock = DockStyle.Fill,
             // Match LogExpert's document style; DockingMdi required test form to be an MDI container.
-			// Using DockingWindow both matches production behavior and allows the test to create real tabs without setting up an artificial MDI container.
+            // Use DockingWindow to match production and avoid requiring
+            // an artificial MDI container in this test.
             DocumentStyle = DocumentStyle.DockingWindow,
             Theme = new VS2015LightTheme()
         };
@@ -218,32 +219,7 @@ internal class TabControllerTests : IDisposable
 
     #endregion
 
-    #region Helpers
-
-    private static LogWindow CreateLogWindow (string fileName)
-    {
-        var coordinatorMock = new Mock<ILogWindowCoordinator>();
-        _ = coordinatorMock.Setup(coordinator => coordinator.ResolveHighlightGroup(It.IsAny<string?>(), It.IsAny<string?>())).Returns(new HighlightGroup());
-        _ = coordinatorMock.SetupGet(coordinator => coordinator.SearchParams).Returns(new SearchParams());
-
-        var configManagerMock = new Mock<IConfigManager>();
-        _ = configManagerMock.SetupGet(configManager => configManager.Settings).Returns(new Settings());
-
-        var pluginRegistryMock = new Mock<IPluginRegistry>();
-        _ = pluginRegistryMock.SetupGet(pluginRegistry => pluginRegistry.RegisteredColumnizers).Returns([new DefaultLogfileColumnizer()]);
-
-        return new LogWindow(
-            coordinatorMock.Object,
-            fileName,
-            isTempFile: false,
-            forcePersistenceLoading: false,
-            configManagerMock.Object,
-            pluginRegistryMock.Object);
-    }
-
-    #endregion
-    
-	#region GetAllWindows Tests
+    #region GetAllWindows Tests
 
     [Test]
     public void GetAllWindows_WhenEmpty_ReturnsEmptyList ()
@@ -365,8 +341,7 @@ internal class TabControllerTests : IDisposable
         // Arrange
         using var controller = new TabController();
 
-       // CreateLogWindow can now provide a non-null LogWindow when needed.
-       // This test verifies that null argument validation happens before the initialization check.
+        // Validate the argument before checking whether the controller is initialized.
         var ex = Assert.Throws<ArgumentNullException>(() => controller.AddWindow(null, "Test"));
         Assert.That(ex.ParamName, Is.EqualTo("window"));
     }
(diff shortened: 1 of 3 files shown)


Hirogen:
looks good, but pls change your indent style from tab to spaces. 
Pls also add a unit test for this case [TabControllerTests.cs] should be the right class

Also change the inner for to an foreach, the for is not needed

Pr0metheus2:
Ok

Pr0metheus2:
The indexed loop for DisplayingContents should be kept; it exposes displayed entries through its custom Count and indexer, while foreach uses the inherited ReadOnlyCollection enumerator and does not return the displayed tabs. The prior LINQ implementation had this failure and caused an empty LastOpenFilesList at shutdown.

src/LogExpert.UI/Services/TabControllerService/TabController.cs
```diff
@@ -461,12 +461,27 @@ protected virtual void Dispose (bool disposing)
...
-                .ToList()
-                .AsReadOnly();
+        if (!_initialized || _dockPanel == null)
+        {
+            return [];
+        }
+
+        List<LogWindow> windows = [];
+
+        foreach (DockPane pane in _dockPanel.Panes)
```
Hirogen, line 471: use var

src/LogExpert.UI/Services/TabControllerService/TabController.cs
```diff
@@ -461,12 +461,27 @@ protected virtual void Dispose (bool disposing)
...
+        {
+            var displayingContents = pane.DisplayingContents;
+
+            for (int index = 0; index < displayingContents.Count; index++)
```
Hirogen, line 477: please add a comment why the for and not the foreach so future developers know why the for is necessary

src/LogExpert.Tests/Services/TabControllerTests.cs
```diff
@@ -191,8 +198,48 @@ public void GetAllWindowsFromDockPanel_ReturnsReadOnlyList ()
...
+        // Act
+        var result = _tabController.GetAllWindowsFromDockPanel();
+
+        // Assert
+        Assert.That(result, Is.EqualTo(new[] { firstWindow, secondWindow, thirdWindow }));
+    }
+
     #endregion
 
+    private static LogWindow CreateLogWindow (string fileName)
```
Hirogen, line 223: add #region Helpers here for the helper

src/LogExpert.Tests/Services/TabControllerTests.cs
```diff
@@ -36,7 +42,8 @@ public void Setup ()
         _dockPanel = new DockPanel
         {
             Dock = DockStyle.Fill,
-            DocumentStyle = DocumentStyle.DockingMdi
+            DocumentStyle = DocumentStyle.DockingWindow,
```
Hirogen, line 44: add a comment why the change from DockingMdi to DockingWindow, for future readers

src/LogExpert.Tests/Services/TabControllerTests.cs
```diff
@@ -36,7 +42,8 @@ public void Setup ()
...
+            Theme = new VS2015LightTheme()
```
Hirogen, line 45: if the theme is necessary in a test project, add the package references so it does not transitive move into the project, but explicitly

src/LogExpert.Tests/Services/TabControllerTests.cs
Hirogen: since there is a new helper CreateLogWindow, this comment needs updating pls update accordingly

Hirogen: this is also no longer correct, pls update the comment accordingly

Hirogen (changes requested):
As a side note, pls also add to the pr description:
this fixes also 2 other places "Close other tabs" [LogTabWindow.cs:2326] and session-save file list [LogTabWindow.cs:2410], both places are affected by this fix and should now work correctly

Pr0metheus2:
All should be fixed as required.

src/LogExpert.Tests/Services/TabControllerTests.cs
Pr0metheus2: added in `src/LogExpert.Tests/LogExpert.Tests.csproj:32~33`

src/LogExpert.Tests/Services/TabControllerTests.cs
Pr0metheus2: Updated `src/LogExpert.Tests/Services/TabControllerTests.cs:18~22`

Pr0metheus2: Updated `src/LogExpert.Tests/Services/TabControllerTests.cs:368~369`

src/LogExpert.UI/Services/TabControllerService/TabController.cs
Pr0metheus2: Done

src/LogExpert.UI/Services/TabControllerService/TabController.cs
Pr0metheus2: done

src/LogExpert.Tests/Services/TabControllerTests.cs
Pr0metheus2: done

src/LogExpert.Tests/Services/TabControllerTests.cs
Pr0metheus2: done

Hirogen:
Sorry been on holidays, here is another review, only /t tabs, pls use spaces

LogExpert.Tests.csproj
-  literal tab. +\t is tab-indented while the PackageReference line below it and every sibling use 4 spaces. 

TabControllerTests.cs
-  literal tab + whitespace-only line. The #region GetAllWindows Tests line is re-emitted as +\t#region GetAllWindows Tests, preceded by a new line containing only four spaces. Same editorconfig rules; the tab also breaks alignment with all 30+ other #region lines in the file.
- mixed indentation inside the initializer. In Setup() the two-line comment splits styles: // Match LogExpert's document style; … (spaces) then \t\t\t// Using DockingWindow both matches production behavior … (tabs). The second line is also ~180 chars where the first was wrapped.
- "CreateLogWindow can now provide a non-null LogWindow when needed" describes a capability the test never exercises; it explains the helper, not the assertion.
- #region Helpers is wedged between GetAllWindowsFromDockPanel Tests and GetAllWindows Tests, breaking the file's one-region-per-SUT-member ordering. Move helpers to the end.

TabController.cs
- misaligned comment. The two // Use 'for' instead of 'foreach': … lines sit at 8 spaces inside a 12-space foreach body, directly above the for they document.

Hirogen (changes requested):
see comment before

Pr0metheus2:
Check please

Hirogen (approved):

PR #697, opened 2026-08-26
Fix/MultiFile pattern (Issue #696)

### LINKED ISSUE #696: Feature Request: Support rotated filenames like app.1.log in MultiFile pattern
#### **Problem**
LogExpert currently only groups MultiFile chains if the index/sequence number is at the **very end** of the filename (e.g., `app.log`, `app.log.1`, `app.log.2`).

When rotated logs put the index **before** the extension (e.g., `app.log`, `app.1.log`, `app.2.log`), LogExpert fails to group them. Patterns like `*$J(.).log` or `*$J(.)` fail and only match `app.log`, `app.log.1`, `app.log.2` etc.

---

#### **Proposed Solution**
Allow index tags (`$J`, `$I`) to appear before a fixed file extension suffix in the **File name pattern** mask.

**Example Pattern:** `*$J(.).log`

* `app.log` ➔ Active log (Index 0, prefix + index hidden)
* `app.1.log` ➔ Rotated log (Index 1)
* `app.2.log` ➔ Rotated log (Index 2)

LogExpert should automatically combine them into a single MultiFile sequence.

---

#### **Current Pattern Syntax Tested**
Based on the application's supported syntax:
* `*` = Any characters (wildcard)
* `$D(<date>)` = Date pattern (`DD`, `MM`, `YY[YY]`)
* `$I` = File index number
* `$J` = File index number, hidden when zero
* `$J(<prefix>)` = Like `$J`, but adding `<prefix>` when non-zero

#### **Why this is needed**
Standard logging frameworks (such as log4net, NLog, log4j, and Linux `logrotate`) commonly output mid-filename index rotation (`app.1.log`). Users cannot always modify third-party log output formats.

Pr0metheus2:
This is a fix for #696

- Fixed/MultiFile rollover patterns with an index before the extension:
    - RolloverFilenameBuilder: the parser was dropping literal text after the final placeholder, including .log.
    - `*$J(.).log` now correctly expands: app.log → app.1.log → app.2.log.

- Updated MultiFile mask dialog legend:
    - Fixed incorrectly displayed `<date>` and `<prefix>` placeholders in English and German resources.
    - Added both usage examples: 
        - `*$J(.)` → app.log, app.log.1, app.log.2 
        - `*$J(.).log` → app.log, app.1.log, app.2.log
    - Added the examples in English, German, and Chinese.
    - Enlarged the dialog and syntax-help label so the date-format lines are no longer clipped.

Before
Legend_old

After
Legend_new

- Fixed/Items in menu File/Multifile -> "Enable MultiFile" and "File name mask...", were always greyed out. Now they are displayed normally after the .log file is open.
    - removed the BackColor and ForeColor assignments for both:
        - multiFileEnabledStripMenuItem
        - multifileMaskToolStripMenuItem

Before
Multifile_old

After
Multifile_new

- Added regression coverage:
    - Filename-builder tests for app.1.log and app.2.log.
    - Reader-level test that loads app.log, app.1.log, and app.2.log in correct order using `*$J(.).log`.
    - Kept the existing `*$J(.)` coverage for app.log.1 / app.log.2.

Verification:
	- Full test suite: 1064 passed, 0 failed, 7 skipped.
	- Resource project build: passed with 0 warnings, 0 errors.


src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
```diff
@@ -238,9 +238,10 @@ private string EscapeNonvarRegions (string formatString)
             }
         }
 
+        _ = result.Append(Regex.Escape(segment.ToString()));
```
Hirogen: Duplicated Code — [RolloverFilenameBuilder.cs:241] is the exact shape of the state-0 flush at line 202. A tiny local FlushEscaped() would let both share it

src/LogExpert.UI/Dialogs/LogTabWindow/LogTabWindow.designer.cs
```diff
@@ -269,18 +269,18 @@ private void InitializeComponent ()
...
             multiFileEnabledStripMenuItem.Size = new Size(165, 22);
             multiFileEnabledStripMenuItem.Text = "Enable MultiFile";
             multiFileEnabledStripMenuItem.Click += OnMultiFileEnabledStripMenuItemClick;
             // 
             // multifileMaskToolStripMenuItem
             // 
-            multifileMaskToolStripMenuItem.BackColor = SystemColors.Control;
-            multifileMaskToolStripMenuItem.ForeColor = SystemColors.ControlDarkDark;
+            // multifileMaskToolStripMenuItem.BackColor = SystemColors.Control;
+            // multifileMaskToolStripMenuItem.ForeColor = SystemColors.ControlDarkDark;
```
Hirogen: if those 2 lilnles are responsible for the grayouted look, delete them, otherwise if someone opens the winforms control, the designer will uncomment them and revive the problem!

src/LogExpert.UI/Dialogs/LogTabWindow/LogTabWindow.designer.cs
```diff
@@ -269,18 +269,18 @@ private void InitializeComponent ()
             // 
             // multiFileEnabledStripMenuItem
             // 
-            multiFileEnabledStripMenuItem.BackColor = SystemColors.Control;
+            // multiFileEnabledStripMenuItem.BackColor = SystemColors.Control;
             multiFileEnabledStripMenuItem.CheckOnClick = true;
-            multiFileEnabledStripMenuItem.ForeColor = SystemColors.ControlDarkDark;
+            // multiFileEnabledStripMenuItem.ForeColor = SystemColors.ControlDarkDark;
```
Hirogen: if those 2 lilnles are responsible for the grayouted look, delete them, otherwise if someone opens the winforms control, the designer will uncomment them and revive the problem!

src/LogExpert.Resources/Resources.de.resx
```diff
@@ -1680,12 +1680,16 @@ Ein ausgewähltes Tool erscheint in der Iconbar. Alle anderen verfügbaren Tools
 		<value>Muster syntax:
 
 * = alle Zeichen (wildcard)
-$D(&amp;lt;date&amp;gt;) = Datumsmuster
+$D(&lt;date&gt;) = Datumsmuster
 $I = Dateiindexnummer
 $J = Dateiindexnummer, versteckt wenn 09
-$J(&amp;lt;prefix&amp;gt;) = Wie $J, jedoch wird ein &amp;lt;prefix&amp;gt; hinzugefügt when es nicht 0 ist
+$J(&lt;prefix&gt;) = Wie $J, jedoch wird ein &lt;prefix&gt; hinzugefügt when es nicht 0 ist
```
Hirogen: wenn

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Hirogen: Index parsing from a rotated file is still broken. Spec: "app.1.log → Rotated log (Index 1)". RolloverFilenameHandler calls SetFileName on whatever file the user opens; for *$J(.).log the regex .*(?'index'\d*)\.log has a greedy .* that swallows the digit in app.1.log, so Index parses as 0 and BuildFileName would produce app.1.1.log. Grouping only works when the active file (app.log) is opened. The same flaw pre-exists for *$J(.) with app.log.1, so it's parity rather than a regression — so the index mapping is only partially delivered.

src/LogExpert.Tests/RollingNameTest.cs
```diff
@@ -32,6 +32,7 @@ public void TestFilename1(string expectedResult, string formatString)
     [TestCase("engine.log", "engine1.log","engine$J.log")]
     [TestCase("engine1.log", "engine2.log","engine$J.log")]
     [TestCase("engine.log", "engine.log.1","*$J(.)")]
+    [TestCase("engine.log", "engine.1.log", "*$J(.).log")]
     [TestCase("engine_2010-06-12.log", "engine_2010-06-12.log.1", "*$D(yyyy-MM-dd).log$J(.)")]
```
Hirogen, line 40: $I untested. For "$J, $I" support, the fix itself generalizes, but no test covers $I or $D followed by a trailing literal (e.g. *$D(yyyy-MM-dd)$J(.).log).

Hirogen, line 40: ParseFormatString computes datePos from the original string but edits the escaped one — a literal with regex metachars before $D (e.g. app.$D(yyyy-MM-dd)) corrupts the regex. Not introduced here, but the new trailing-literal support makes such patterns likelier.

Hirogen (changes requested):

src/LogExpert.Resources/Resources.de.resx
Pr0metheus2: Fixed from when to wenn

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: - fixed parsing of `$J(<prefix>)` from rotated files, so both `app.log.1` and `app.1.log` correctly identify index 1.
- made wildcard matching non-greedy and anchored the generated regex to the complete filename.
- updated rebuilding logic to remove an existing prefix plus index together before inserting the new index.
- fixed $D(...) parsing to use positions in the escaped format string, preventing errors when a literal regex character such as . appears before $D

Added coverage for opening:
  - engine.log.1 with `*$J(.)` → builds engine.log.2
  - engine.1.log with `*$J(.).log` → builds engine.2.log

src/LogExpert.Tests/RollingNameTest.cs
Pr0metheus2: Fixed by locating and parsing $D(...) in the escaped format string. Offsets remain correct when literal regex metacharacters appear before the date placeholder. 
Added a regression test for app.$D(yyyy-MM-dd).log that verifies date parsing and rebuilding.

Pr0metheus2: Added cases which cover:
- $I with a trailing literal .log, starting from a rotated file.
- $D plus $J(.) followed by a trailing .log, also starting from a rotated file.

src/LogExpert.UI/Dialogs/LogTabWindow/LogTabWindow.designer.cs
Pr0metheus2: Removed as per request

src/LogExpert.UI/Dialogs/LogTabWindow/LogTabWindow.designer.cs
Pr0metheus2: Removed as per request

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: Added FlushEscaped() to share duplicated escaping code.

Pr0metheus2:
Check please

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
```diff
@@ -192,15 +203,20 @@ private string EscapeNonvarRegions (string formatString)
         StringBuilder result = new();
         StringBuilder segment = new();
 
+        void FlushEscaped ()
+        {
+            _ = result.Append(Regex.Escape(segment.ToString()));
+            segment = new StringBuilder();
+        }
```
Hirogen: extract it to a real function not a function in a function, this only leads to unreadable code

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
```diff
@@ -192,15 +203,20 @@ private string EscapeNonvarRegions (string formatString)
         StringBuilder result = new();
         StringBuilder segment = new();
 
+        void FlushEscaped ()
```
Hirogen: FlushEscaped() covers only case 0, while case 1 and case 3 still inline _ = result.Append(segment); segment = new StringBuilder();.

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
```diff
@@ -116,15 +116,23 @@ public string BuildFileName ()
 
         if (_indexGroup != null && _indexGroup.Success)
         {
-            fileName = fileName.Remove(_indexGroup.Index, _indexGroup.Length);
+            var indexPosition = _indexGroup.Index;
```
Hirogen: Hidden invariant / Feature Envy — indexPosition = _condGroup.Index; indexLength += _condGroup.Length; silently assumes cond is contiguous and immediately precedes index — an invariant set 60 lines away in ParseFormatString. Uncommented; changing either regex breaks the other.

src/LogExpert.Tests/RollingNameTest.cs
```diff
@@ -54,6 +60,16 @@ public void TestFilenameAnd2(string fileName, string expectedResult, string form
         Assert.That(name, Is.EqualTo(expectedResult));
     }
 
+    [Test]
+    public void BuildFileName_DatePatternAfterRegexMetacharacter_IncrementsDate ()
```
Hirogen: BuildFileName_X_Y while the file's existing methods are TestFilenameAnd1/2.

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
```diff
@@ -176,12 +184,15 @@ private void ParseFormatString (string formatString)
             }
         }
 
-        fmt = fmt.Replace("*", ".*", StringComparison.Ordinal);
+        fmt = fmt.Replace("*", ".*?", StringComparison.Ordinal);
         _hideZeroIndex = fmt.Contains("$J", StringComparison.Ordinal);
         fmt = fmt.Replace("$I", "(?'index'[\\d]+)", StringComparison.Ordinal);
-        fmt = fmt.Replace("$J", "(?'index'[\\d]*)", StringComparison.Ordinal);
+        var optionalIndexPattern = _condContent != null
```
Hirogen: optionalIndexPattern reads as "the pattern is optional";

src/LogExpert.Tests/StreamReaderTests/LogfileReaderMultiFileFlagTests.cs
```diff
@@ -26,6 +26,8 @@ public void SetUp ()
 
         File.Copy(Path.Combine(_testDataDirectory, "app.log"), _logFile);
         File.Copy(Path.Combine(_testDataDirectory, "app.log.1"), _logFile + ".1");
+        File.Copy(Path.Combine(_testDataDirectory, "app.1.log"), Path.Combine(_testDirectory, "app.1.log"));
```
Hirogen: SetUp copies app.1.log/app.2.log unconditionally, so all four tests now see two extra files though only the new one needs them.

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
```diff
@@ -176,12 +184,15 @@ private void ParseFormatString (string formatString)
             }
         }
 
-        fmt = fmt.Replace("*", ".*", StringComparison.Ordinal);
+        fmt = fmt.Replace("*", ".*?", StringComparison.Ordinal);
         _hideZeroIndex = fmt.Contains("$J", StringComparison.Ordinal);
         fmt = fmt.Replace("$I", "(?'index'[\\d]+)", StringComparison.Ordinal);
```
Hirogen: "Allow index tags ($J, $I) to appear before a fixed file extension suffix." RolloverFilenameBuilder.cs:189 still maps $I → (?'index'[\d]+); only $J gets the prefix-aware alternation (:190-193). So *$I.log cannot match the active file app.log — SetFileName fails, IsIndexPattern stays false, no chain is built. The one new $I test (RollingNameTest.cs:36, engine1.log→engine2.log) starts from a numbered file and never exercises the failing case.

src/LogExpert.Resources/Resources.de.resx
```diff
@@ -1680,12 +1680,16 @@ Ein ausgewähltes Tool erscheint in der Iconbar. Alle anderen verfügbaren Tools
 		<value>Muster syntax:
 
 * = alle Zeichen (wildcard)
-$D(&amp;lt;date&amp;gt;) = Datumsmuster
+$D(&lt;date&gt;) = Datumsmuster
 $I = Dateiindexnummer
 $J = Dateiindexnummer, versteckt wenn 09
```
Hirogen: still reads $J = Dateiindexnummer, versteckt wenn 09 — should be 0. Correct in the English and Chinese files.

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
```diff
@@ -176,12 +184,15 @@ private void ParseFormatString (string formatString)
             }
         }
 
-        fmt = fmt.Replace("*", ".*", StringComparison.Ordinal);
+        fmt = fmt.Replace("*", ".*?", StringComparison.Ordinal);
```
Hirogen, line 179: .*→.*? and :194 \A…\z anchoring change every mask, not just ones with trailing literals. It holds because RolloverFilenameHandler.cs:39 passes a bare filename, not a path — but that invariant is undocumented and untested; a full path with any non-*-prefixed mask now fails where it previously matched.

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
```diff
@@ -116,15 +116,23 @@ public string BuildFileName ()
```
Hirogen, line 116: A saved mask *$J(.).log previously behaved as *$J(.) and matched app.log.1; it now means app.1.log. That's what the spec wants, but existing user settings change

Hirogen (changes requested):

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: Should the corrected syntax remain this way?
- *$J(.) → app.log, app.log.1, app.log.2
- *$J(.).log → app.log, app.1.log, app.2.log

If so, should I migrate existing saved *$J(.).log masks to *$J(.)?
(7)

src/LogExpert.Resources/Resources.de.resx
Pr0metheus2: Fixed (versteckt wenn null)

src/LogExpert.Tests/StreamReaderTests/LogfileReaderMultiFileFlagTests.cs
Pr0metheus2: Two copies moved to `SingleFileCtor_MultiFileTrue_LoadsIndexBeforeExtension()`.
(9)

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: Moved the helper out of EscapeNonvarRegions() into a private class method. (1)

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: Added two shared helpers: one appends an escaped literal segment and one appends a placeholder segment. Both also clear the segment, so states 0, 1, and 3 now use the same append & clear behavior. (2)

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: Removed the separate prefix-position calculation. The generated regex now captures the complete index part, including its prefix when present, in one indexPart group. Rebuilding removes and replaces that explicit group directly. (3)

src/LogExpert.Tests/RollingNameTest.cs
Pr0metheus2: Renamed to `TestFilenameDateAfterRegexMetacharacter`
(8)

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: Renamed the pattern variables to indexPattern and conditionalIndexPattern to describe what they match rather than suggesting that the whole pattern is optional. (4)

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: Updated $I to match an unnumbered active file as well as numbered rotated files.
Added coverage for app.log becoming app1.log after the index is incremented.
(5)

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: Kept the end anchor needed to parse an index before fixed suffix, but removed the start anchor. 
This retains compatibility with full paths and non-wildcard masks. 
Added a full-path regression test.
(6)

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: Added settings migration for corrected MultiFile mask behavior.

To preserve compatibility, settings version 2 migrates existing saved *$J(.).log preferences to *$J(.). New version-2 settings keep both patterns distinct:
  - *$J(.) for app.log.1
  - *$J(.).log for app.1.log

Added tests for migrating old settings and for ensuring new settings are not changed.

src/LogExpert.Configuration/LegacyPreferencesMigrator.cs
```diff
@@ -55,4 +62,14 @@ private static void MigrateToV1 (Settings settings)
         }
 #pragma warning restore CS0618
     }
+
+    private static void MigrateToV2 (Settings settings)
+    {
+        // Before the parser fix, the trailing .log in this pattern was ignored, making it behave
+        // like *$J(.). Preserve that behavior for existing saved preferences.
```
Hirogen: rewrites a saved *$J(.).log to *$J(.). A user who already saved that pattern hoping for the new grouping, which is exactly the situation, gets silently reverted to the old behaviour. It bumps the schema version for every user and does not touch per-file persisted patterns, so the "preserve old behaviour" rationale is inconsistent.

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
```diff
@@ -176,12 +176,16 @@ private void ParseFormatString (string formatString)
...
-
-        _regex = new Regex(fmt);
+        const string indexPattern = "(?'indexPart'(?'index'[\\d]*))";
+        var conditionalIndexPattern = _condContent != null
+            ? $"(?'indexPart'(?:{Regex.Escape(_condContent)}(?'index'[\\d]+)|(?'index')))"
+            : indexPattern;
+        fmt = fmt.Replace("$I", indexPattern, StringComparison.Ordinal);
+        fmt = fmt.Replace("$J", conditionalIndexPattern, StringComparison.Ordinal);
+
+        _regex = new Regex(fmt + @"\z");
```
Hirogen, line 189: $I digit class went from one-or-more to zero-or-more. A plain app.log now matches *$I.log as index 0, so $I behaves like $J when the opened file has no digits.

*$J(.) no longer behaves exactly as before. The lazy wildcard plus the end anchor mean opening engine.log.1 now yields index 1. Previously it yielded index 0 and generated engine.log.1.1. This is a fix, but it is an unstated behaviour change.

The FlushEscaped method named in the PR description does not exist. The helpers are AppendEscapedSegment and AppendSegment.

src/LogExpert.Tests/StreamReaderTests/LogfileReaderMultiFileFlagTests.cs
```diff
@@ -68,6 +68,24 @@ public void SingleFileCtor_MultiFileTrue_ExpandsRollover ()
         });
     }
 
+    [Test]
+    public void SingleFileCtor_MultiFileTrue_LoadsIndexBeforeExtension ()
+    {
+        var options = new MultiFileOptions { FormatPattern = "*$J(.).log" };
```
Hirogen, line 74: The $I case is only tested from the active file, despite the PR description claiming a rotated-file $I test.

src/LogExpert.Tests/StreamReaderTests/LogfileReaderMultiFileFlagTests.cs
```diff
@@ -91,15 +109,15 @@ public void MultiFileCtor_AlwaysMultiFile ()
         });
     }
 
-    private LogfileReader CreateSingleFileReader (bool multiFile)
+    private LogfileReader CreateSingleFileReader (bool multiFile, MultiFileOptions? options = null)
     {
         return new LogfileReader(
             _logFile,
             new EncodingOptions { Encoding = Encoding.UTF8 },
```
Hirogen, line 116: opens the active file only. Opening a rotated file directly never picks up lower siblings because the name-list walk only goes upward. That is pre-existing, but the spec is only met when the active file is opened.

No round-trip test for *$J(.).log.

Hirogen (changes requested):

src/LogExpert.Tests/StreamReaderTests/LogfileReaderMultiFileFlagTests.cs
Pr0metheus2: My understanding is that this is intended behavior: when a rotated file such as app.1.log is opened, MultiFile loads that file and higher indexes only. To load the complete chain including lower indexes, the user should open the active file (app.log). If user want to open log file from specified date only and further, i.e. app.8.log, this is a useful behaviour and in this case app will load app.8.log -> app.9.log etc.

Please confirm whether this is correct. If MultiFile should also discover lower-index siblings when a rotated file is opened directly, I can extend the handler and add coverage for that behavior.

src/LogExpert.Configuration/LegacyPreferencesMigrator.cs
Pr0metheus2: (1) OK, I've removed the settings migration completely:
- Schema version restored to 1.
- Removed MigrateToV2().
- Removed its tests and test-only import.
- Existing saved *$J(.).log masks now retain the corrected new meaning.

src/LogExpert.Core/Classes/Log/RolloverFilenameBuilder.cs
Pr0metheus2: (2) Restored $I behavior:
- $I again requires one or more digits ([\d]+).
- app.log no longer matches *$I.log.
- $J retains its zero-index handling and the rotated-file parsing fix.
- The corrected behavior for opening engine.log.1 is kept; describe that explicitly in the PR rather than treating it as invisible behavior.

Note: I cannot find FlushEscaped in the current PR description.

src/LogExpert.Tests/StreamReaderTests/LogfileReaderMultiFileFlagTests.cs
Pr0metheus2: (3) Updated the $I test to start from a rotated filename:
engine1.log → engine2.log 
with *$I.log.

Hirogen:
sorry for longer not writing, got a lot todo at work, just one more thing:

^^

In the dialog layout: at 100% scaling, the German help text needs 270 pixels of height, but the label allows 254, so the final date-format line is still clipped.

after that, we can merged it

Pr0metheus2:
Fixed the German 100%-scaling clipping to 270px of height.

Hirogen (approved):

